PR #6, opened 2020-02-03
Basic configuration

tisnik:
# Description

Basic configuration file + its loading.

## Type of change

## Testing steps

Run `make test`


config.toml
```diff
@@ -0,0 +1,12 @@
+[broker]
+address = "localhost:9092"
+topic = "platform.results.ccx"
```
quarckster: Topic name is `ccx.ocp.results`

tisnik: nice catch, I've fixed it

PR #20, opened 2020-02-06
created tests for storage using sqlite in memory as a way to mock db

### LINKED ISSUE #18: Incorporate DB mocks into unit tests
Worth to view and check:
https://github.com/DATA-DOG/go-sqlmock

Sergey1011010:
# Description

Created tests for storage using sqlite in memory as a way to mock db

Fixes #18 

## Type of change

## Testing steps

`make test` or `go test ./... -run "TestMockDBStorage.*"`

## Notes

https://github.com/DATA-DOG/go-sqlmock doesn't fit very well because it requires us to call methods like `ExpectQuery("SELECT something FROM somewhere .*")` for each request our storage methods are doing and I think that unit tests should not be aware of internal implementation of the storage. 

## Questions

1. Should the `cluster` field be unique? Because now if we are pushing two rows `orgID: 1, cluster: "cluster name"` and `orgID: 2, cluster: "cluster name"`  (with the same cluster field) we will end up having only second row in database

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -47,8 +47,8 @@ func TestMockDBStorageReadReport(t *testing.T) {
 	defer mockStorage.Close()
 
 	const testOrgID = storage.OrgID(1)
-	const testClusterName = storage.ClusterName("cluster name")
-	const testClusterReport = storage.ClusterReport("cluster report")
+	const testClusterName = storage.ClusterName("84f7eedc-0dd8-49cd-9d4d-f6646df3a5bc")
+	const testClusterReport = storage.ClusterReport("{}")
 
 	err = mockStorage.WriteReportForCluster(testOrgID, testClusterName, testClusterReport)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


storage/storage_test.go
```diff
@@ -15,3 +15,112 @@ limitations under the License.
...
+	"testing"
+
+	"github.com/RedHatInsights/insights-results-aggregator/storage"
+	"github.com/stretchr/testify/assert"
+)
+
+func getMockStorage() (storage.Storage, error) {
+	mockStorage, err := storage.New(storage.Configuration{
+		Driver:     "sqlite3",
+		DataSource: ":memory:",
```
tisnik: oh this is THE trick :) good to know

storage/storage_test.go
```diff
@@ -15,3 +15,112 @@ limitations under the License.
...
+}
+
+func TestMockDBStorageListOfClustersForOrg(t *testing.T) {
+	mockStorage, err := getMockStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer mockStorage.Close()
+
+	err = mockStorage.WriteReportForCluster(1, "cluster 1", "report 1")
```
tisnik: it works right now, but in the future we need to check if the cluster name is in "GUID"-like format. https://en.wikipedia.org/wiki/Universally_unique_identifier So I'd add that cluster name instead of 'cluster 1'.
Also (it was not mentioned, my fault) - report needs to be in JSON format. It would need to be checked later as well.

PR #22, opened 2020-02-06
Server stub

tisnik:
# Description

Stub for HTTP server.

## Type of change

## Testing steps
Use `make test` as usual please.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -16,7 +16,12 @@ limitations under the License.
 
 package server
 
-import ()
+import (
+	"github.com/gorilla/mux"
+	"io"
+	"log"
+	"net/http"
+)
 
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -30,10 +35,50 @@ type Impl struct {
...
+
+	if err != nil {
+		log.Fatal("Unable to initialize HTTP server", err)
+	}
+	return nil
 }
 
 // Start starts server
 func (server Impl) Start() error {
+	server.Initialize()
```
Sergey1011010: Shouldn't the error be handled here? It does not return any errors as I can see, but someone may add code returning errors to `Initialize` at some point

PR #33, opened 2020-02-10
Consumer tests

tisnik:
# Description

More thorough unit tests for Consumer interface.

## Type of change

## Testing steps
Please run `make test` as usual.

### DIFF
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -100,7 +100,10 @@ func (consumer Impl) Start() error {
 	consumed := 0
 	for {
 		msg := <-consumer.PartitionConsumer.Messages()
-		consumer.ProcessMessage(msg)
+		err := consumer.ProcessMessage(msg)
+		if err != nil {
+			log.Println("Error processing message consumed from Kafka:", err)
+		}
 		consumed++
 	}
 }
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -31,6 +31,7 @@ import (
 type Consumer interface {
 	Start() error
 	Close() error
+	ProcessMessage(msg *sarama.ConsumerMessage)
 }
 
 // Impl in an implementation of Consumer interface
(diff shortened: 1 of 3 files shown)


consumer/consumer.go
```diff
@@ -102,17 +100,24 @@ func (consumer Impl) Start() error {
...
-		}
-		err = consumer.Storage.WriteReportForCluster(orgID, clusterName, report)
-		if err != nil {
-			log.Println("Error writing report to database:", err)
-		}
+	}
+}
+
+// ProcessMessage processes an incoming message
+func (consumer Impl) ProcessMessage(msg *sarama.ConsumerMessage) {
```
JoseLSegura: As totally newbie point of view:
Why don't returning an `error` too? When an `err` is returned from the parsing of the message, you can "inform" about it to the caller

consumer/consumer_test.go
```diff
@@ -47,3 +49,164 @@ func TestConsumerConstructorNoKafka(t *testing.T) {
...
+	}
+}
+
+func TestParseMessageWithImproperJSON(t *testing.T) {
+	const message = `"this_is_not_json_dude"`
+	_, _, _, err := consumer.ParseMessage([]byte(message))
+	if err == nil {
+		t.Fatal("Error is expected to be returned for empty message")
+	}
+}
```
JoseLSegura: Can the `err` be checked in order to see if the error is the expected one?

consumer/consumer_test.go
```diff
@@ -47,3 +49,164 @@ func TestConsumerConstructorNoKafka(t *testing.T) {
...
+
+func TestParseMessageWithWrongContent(t *testing.T) {
+	const message = `{"this":"is", "not":"expected content"}`
+	_, _, _, err := consumer.ParseMessage([]byte(message))
+	if err == nil {
+		t.Fatal("Error is expected to be returned for empty message")
```
JoseLSegura: This error string can be confusing, because the `message` is incorrect, but not empty

tisnik: oh good catch!

consumer/consumer_test.go
tisnik: you are right, I'll fix it.

consumer/consumer.go
tisnik: yes, we can do it. thanks for review!

JoseLSegura: :-) Glad to help

consumer/consumer.go
```diff
@@ -102,20 +100,29 @@ func (consumer Impl) Start() error {
 	consumed := 0
 	for {
 		msg := <-consumer.PartitionConsumer.Messages()
-		log.Printf("Consumed message offset %d\n", msg.Offset)
+		consumer.ProcessMessage(msg)
```
JoseLSegura: Probably you want to check if `ProcessMessage` returned an error

tisnik: fixed, nice catch!

PR #36, opened 2020-02-11
Added Dockerfile

quarckster:
# Description

Added a dockerfile for building the application.

## Type of change

## Testing steps
I ran `podman build -f Dockerfile -t ira .`  locally and the image was built successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM golang:1.13 AS builder
+
+COPY . insights-results-aggregator
+
+RUN cd insights-results-aggregator && \
+    make build
+
+FROM registry.access.redhat.com/ubi8-minimal
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+FROM registry.access.redhat.com/ubi8-minimal
+
+COPY --from=builder /go/insights-results-aggregator/insights-results-aggregator .
+
+RUN chmod a+x /insights-results-aggregator
+
+USER 1001
+
+CMD ["/insights-results-aggregator"]
```
bond95: You should define mode ("consumer" or "server") in arguments, could it be as ENV var?

tisnik: it will be 'fixed' in https://github.com/RedHatInsights/insights-results-aggregator/pull/37

quarckster: @tisnik thank you!

PR #81, opened 2020-02-14
Adding a check for the ClusterName UUID format

### LINKED ISSUE #68: Check if the cluster name read from Kafka topic is correct GUID
Possibly useful libraries to use:
https://github.com/satori/go.uuid
https://godoc.org/github.com/google/uuid

JoseLSegura:
# Description

Checks that the received ClusterName is a proper UUID using the  github.com/google/uuid package.
If it is not an UUID, the processMessage will fail and an adequate error reported

Fixes #68 

## Type of change

## Testing steps
make default test

### DIFF
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -24,6 +24,7 @@ import (
 	"log"
 
 	"github.com/Shopify/sarama"
+	"github.com/google/uuid"
 
 	"github.com/RedHatInsights/insights-results-aggregator/broker"
 	"github.com/RedHatInsights/insights-results-aggregator/metrics"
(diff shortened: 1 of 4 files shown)


JoseLSegura:
I only perform the check. I was trying to change the `types.ClusterName` from `string` to `uuid.UUID`, but I found a lot of problems with that. The hardest one is about `json.Unmarshal`, that always parse JSON strings to Go strings, and I didn't find a way to try to covert to a custom type

JoseLSegura:
> Cool, thank you. Also I think it will be nice to add validation of identity on pipeline side

I agree, I'll create the issue there too

JoseLSegura:
Oh, I just see that you have created it :-D

PR #88, opened 2020-02-14
added last checked timestamp

### LINKED ISSUE #44: New feature request: ability to display timestamp when insights data was 'last checked'
It is needed to display the 'last checked' field for given cluster on UI, so we need to be able to store timestamp in a storage.

Sergey1011010:
# Description

added last checked timestamp

Fixes #44 

## Type of change

## Testing steps

`make test` & `make integration_tests`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ cyclo: ## Run gocyclo
 style: fmt vet lint cyclo ## Run all the formatting related commands (fmt, vet, lint, cyclo)
 
 run: clean build ## Build the project and executes the binary
-	./insights-operator-aggregator 
+	./insights-results-aggregator
 
 test: clean build ## Run the unit tests
 	@go test -coverprofile coverage.out $(shell go list ./... | grep -v tests)
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -19,6 +19,7 @@ package storage_test
 import (
 	"github.com/stretchr/testify/assert"
 	"testing"
+	"time"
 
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ cyclo: ## Run gocyclo
 style: fmt vet lint cyclo ## Run all the formatting related commands (fmt, vet, lint, cyclo)
 
 run: clean build ## Build the project and executes the binary
-	./insights-operator-aggregator 
+	./insights-results-aggregator
```
tisnik: good catch

storage/storage_test.go
```diff
@@ -57,7 +58,7 @@ func checkReportForCluster(t *testing.T, storage storage.Storage, orgID types.Or
 }
 
 func writeReportForCluster(t *testing.T, storage storage.Storage, orgID types.OrgID, clusterName types.ClusterName, report types.ClusterReport) {
-	err := storage.WriteReportForCluster(orgID, clusterName, report)
+	err := storage.WriteReportForCluster(orgID, clusterName, report, time.Now())
```
tisnik: not sure whether you want to `time.Now().UTC()`

tisnik:
@Sergey1011010 prob. it might be worth to update scripts in `local_storage`

storage/storage_test.go
Sergey1011010: looks like it won't matter because we will have timestamp in db, but I'm not sure too

tisnik: I see, TY

PR #89, opened 2020-02-14
Adding check UUID format for cluster name in HTTP requests

### LINKED ISSUE #69: Check if cluster name in REST API requests is correct GUID
Possibly useful libraries to use:
https://github.com/satori/go.uuid
https://godoc.org/github.com/google/uuid

JoseLSegura:
# Description

Checking UUID format on the cluster name for the get reports HTTP requests.

I added some unit tests for the server package. IDK if it is aligned with the rest of the unit tests or even if it is well done because this are my first unit tests in Go

Fixes #69 

## Type of change

## Testing steps
Usual "make test"

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -47,6 +47,7 @@ import (
 	"github.com/RedHatInsights/insights-results-aggregator/metrics"
 	"github.com/RedHatInsights/insights-results-aggregator/storage"
 	"github.com/RedHatInsights/insights-results-aggregator/types"
+	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -131,7 +132,13 @@ func (server HTTPServer) readClusterName(writer http.ResponseWriter, request *ht
 		responses.SendInternalServerError(writer, message)
 		return types.ClusterName(""), errors.New(message)
 	}
-	// TODO: add check for GUID-like name
+
+	if _, parseErr := uuid.Parse(clusterName); parseErr != nil {
```
tisnik: just `err` is IMHO more idiomatic in Go

server/server.go
```diff
@@ -151,7 +158,7 @@ func (server HTTPServer) listOfClustersForOrganization(writer http.ResponseWrite
 	}
 }
 
-func (server HTTPServer) readReportForCluster(writer http.ResponseWriter, request *http.Request) {
+func (server HTTPServer) ReadReportForCluster(writer http.ResponseWriter, request *http.Request) {
```
tisnik: this need not be exported (ie. publicly visible)

tisnik: and `goling` basically said the same on CI :)

JoseLSegura: Oh, sorry. I pushed it by mistake. Initially I was using a different (and not productive at all) approach for the UT and I tried to use this function directly, so I needed to change the visibility and forget to get to its correct visibility

server/server.go
JoseLSegura: Ok

Issue #90, opened 2020-02-17
Story: Auth. scheme for accessing Insigts results from OCM


tisnik:
As an user I expect I'll be able to be authorized and authenticated so I'll be sure that users from different organizations won't see results for clusters in my organization.

bond95:
If I understand whole mechanism right, auth can be done through `x-rh-identity` which contains all needed information including `org_id` and `is_org_admin`

tisnik:
It might be correct @bond95 , but we also need to check if the user is in the correct org id to get the data for given cluster.

bond95:
@tisnik already done in https://github.com/RedHatInsights/insights-results-aggregator/pull/94

tisnik:
I mean - let's say I'm user from organization X so I should not be able to see anything about clusters from organization Y.

bond95:
> I mean - let's say I'm user from organization X so I should not be able to see anything about clusters from organization Y.

Yes, now it's working like this.

tisnik:
Oh ok then. I must missed it in the PR.

tisnik:
Is it possible pls. to have an example how it works for all three endpoints we have currently? TYVM!

PR #94, opened 2020-02-18
Added support for x-rh-identity

bond95:
# Description

Added support for x-rh-identity, but only for `production` mode

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/90

## Type of change

## Testing steps
Set in server configuration `Debug` to `false` and use x-rh-identity header

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -1,0 +1,82 @@
+// Auth implementation based on JWT
+
+/*
+Copyright © 2019, 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"github.com/RedHatInsights/insights-operator-utils/responses"
+	"net/http"
+)
+
+type contextKey string
+
+const (
+	contextKeyUser = contextKey("user")
+)
+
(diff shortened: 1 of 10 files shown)


bond95:
@tisnik WDYT?

server/auth.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/http"
+)
+
+type contextKey string
+
+const (
+	contextKeyUser = contextKey("user")
+)
+
+type Internal struct {
```
tisnik: if this is exported, it needs to have proper docstring

server/auth.go
```diff
@@ -0,0 +1,80 @@
...
+	OrgId string `json:"org_id"`
+}
+
+type Identity struct {
```
tisnik: dtto

server/server.go
```diff
@@ -53,6 +54,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var Environment = os.Getenv("AGGREGATOR_ENV")
```
tisnik: I'd urge to use Configuration there, not "package global" variables. Please

server/server.go
```diff
@@ -111,6 +117,16 @@ func (server HTTPServer) readOrganizationID(writer http.ResponseWriter, request
 		return 0, errors.New(message)
 	}
 
+	if identityContext != nil && Environment == "production" {
+		identity := identityContext.(Identity)
+		if Environment == "production" && identity.Internal.OrgId != organizationIDParam {
```
tisnik: enum type there?

server/auth.go
tisnik: IMHO it might not be exported...?

bond95:
@tisnik @natiiix thank you all for feedback

server/server.go
tisnik: for example we can add "production bool" into the Configuration. Do the config setup properly in main package and let be server itself basically global-variable free. WDYT?

server/server.go
tisnik: or better: please add "production bool" into Configuration and change this line to something like: if server.config.production && ...

tisnik:
just please make CI happy :)

PR #98, opened 2020-02-18
Change list-returning endpoints to plural

### LINKED ISSUE #96: Organizations endpoint should be plural
Collections should be in plural from. https://github.com/RedHatInsights/insights-results-aggregator/blob/master/openapi.yml#L9 should be changed to `/organizations:`

natiiix:
# Description

Modified endpoint names to match the requirements.

The tests all seem to pass, but I assume this change will break something.

Fixes #96 

## Type of change

## Testing steps

❓

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -18,9 +18,9 @@ limitations under the License.
 // Insights results aggregator service. In current version, the following
 // REST API endpoints are available:
 //
-// API_PREFIX/organization - list of all organizations (HTTP GET)
+// API_PREFIX/organizations - list of all organizations (HTTP GET)
 //
-// API_PREFIX/cluster/{organization} - list of all clusters for given organizations (HTTP GET)
+// API_PREFIX/organizations/{organization}/clusters - list of all clusters for given organization (HTTP GET)
 //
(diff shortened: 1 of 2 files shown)


natiiix:
@tisnik Oops, didn't notice that.

server/server.go
```diff
@@ -18,9 +18,9 @@ limitations under the License.
 // Insights results aggregator service. In current version, the following
 // REST API endpoints are available:
 //
-// API_PREFIX/organization - list of all organizations (HTTP GET)
+// API_PREFIX/organizations - list of all organizations (HTTP GET)
 //
-// API_PREFIX/cluster/{organization} - list of all clusters for given organizations (HTTP GET)
```
quarckster: > list of all clusters for given organizations (HTTP GET)
Does it mean that I can provide more than one orgID separated by a comma and receive a list of clusters?

tisnik: nope, should be organizations/organization

natiiix:
> Basically looks good, but I think we need to update `report` request as well, to be a part of `organizations/{organization}/clusters` chain

@bond95 Not sure what you mean, but we cannot do `/organizations/{orgId}/cluters/{clusterId}/report`.

PR #99, opened 2020-02-18
Unit tests

### LINKED ISSUE #86: Unit test coverage must exceed a promotion gate of 85%
This is a requirement for onboard application to cloud.redhat.com

Sergey1011010:
# Description

Covered most of the code with tests

Fixes #86 and #112 

## Type of change

## Testing steps

`make test` & `make integration_tests`


server/server.go
```diff
@@ -78,18 +77,18 @@ func logRequestHandler(writer http.ResponseWriter, request *http.Request, nextHa
 }
 
 // LogRequest - middleware for loging requests
-func (server HTTPServer) LogRequest(nextHandler http.Handler) http.Handler {
+func (server *HTTPServer) LogRequest(nextHandler http.Handler) http.Handler {
```
tisnik: is it a form of (preliminary) optimization?

tests/helpers/mock_storage.go
```diff
@@ -0,0 +1,39 @@
...
+			return nil, err
+		}
+	}
+
+	return mockStorage, nil
+}
+
+// MustGetMockStorage creates mocked storage based on in-memory Sqlite instance
+// produces t.Fatal(err) on error
+func MustGetMockStorage(t *testing.T, init bool) storage.Storage {
```
tisnik: i like this approach!

aggregator.go
```diff
@@ -41,6 +43,11 @@ const (
 	ExitStatusServerError
 )
 
+var (
+	serverInstance   *server.HTTPServer = nil
```
tisnik: probably linters wont be happy with "= nil" part as it is redundant

tisnik:
@Sergey1011010 are you expecting that the real Kafka broker is started in CI during unit testing? It is not the case right now.

server/server.go
Sergey1011010: I needed to modify server instance in `Start()` https://github.com/Sergey1011010/insights-results-aggregator/blob/unit-tests/server/server.go#L165 to be able to shut it down later and changed every method according to style guides(should be either pointer or value for every method in struct)

tisnik: ah ok, I see. TY (for me, if receiver is pointer, it 'says' that the method is about to modify receiver, but dunno, let's use your approach)

aggregator.go
```diff
@@ -108,6 +115,27 @@ func startService() {
...
+			break
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+}
+
+func stopService() {
+	err := serverInstance.Stop(context.TODO())
+	if err != nil {
+		log.Fatal(err)
```
tisnik: hmm in this case it might be probably more correct to try to Close() consumer before log.Fatal. WDYT?

server/router_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/google/uuid"
+	"github.com/gorilla/mux"
+)
+
+// RouterMissingParamError missing parameter in URL
+type RouterMissingParamError struct {
+	paramName string
+}
+
+func (e *RouterMissingParamError) Error() string { return fmt.Sprintf("missing param %v", e.paramName) }
```
tisnik: formatting?

server/router_utils.go
```diff
@@ -0,0 +1,126 @@
...
+			paramName: paramName, paramValue: value, errString: "integer expected",
+		}
+	}
+
+	return intValue, nil
+}
+
+// GetRouterPositiveIntParam retrieves parameter from URL like `/organization/{org_id}`
+// and check it for being valid and positive integer, otherwise returns error
+func GetRouterPositiveIntParam(request *http.Request, paramName string) (int64, error) {
```
bond95: Is this should be exported?

server/router_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	if !found {
+		return "", &RouterMissingParamError{paramName: paramName}
+	}
+
+	return value, nil
+}
+
+// GetRouterIntParam retrieves parameter from URL like `/organization/{org_id}`
+// and check it for being valid integer, otherwise returns error
+func GetRouterIntParam(request *http.Request, paramName string) (int64, error) {
```
bond95: Is this should be exported? Because it used only in package `server`

server/router_utils.go
tisnik: (+ positive int -> uint/uint64)

aggregator.go
Sergey1011010: fixed

aggregator.go
Sergey1011010: fixed? https://github.com/RedHatInsights/insights-results-aggregator/pull/99/commits/712348e379fccd08bf4e6067417da165a09bc5fe

server/router_utils.go
Sergey1011010: What do you mean?

server/router_utils.go
Sergey1011010: @bond95 fixed

server/router_utils.go
Sergey1011010: @bond95 fixed

server/router_utils.go
Sergey1011010: @tisnik the main purpose of that was to be compatible with `readOrganizationID` which returns `types.OrgID` which is int and to check for invalid IDs(negative and 0). Maybe, we could also change `types.OrgID` to uint64 since there's no databases with negative ids?

server/router_utils.go
natiiix: @Sergey1011010 It's rather unconventional to have an actual one-liner function in Go. I belive that's what Pavel meant here.

Sergey1011010: @natiiix sounds reasonable, fixed

server/router_utils.go
tisnik: yes, that might be better

bond95:
@Sergey1011010 do rebase, please.

server/router_utils.go
Sergey1011010: fixed

Sergey1011010:
@tisnik how do I trigger CI after last 2 commits?

Sergey1011010:
rebase, again.. @bond95 thanks

PR #100, opened 2020-02-18
Implement DB migration logic

### LINKED ISSUE #93: Create mechanism for DB migration
Project now don't support DB migration, which makes a lot of problem on deploy. So we need to create mechanism for migrations that will check and update DB any time it's starting.

natiiix:
# Description

I added some code for DB migrations. There are also tests for the migration logic, but they only cover the bare minimum at the moment.

- Version 0 represents an empty database.
- To get to any other version, one must first initialize the DB migration table and then upgrade the database to the desired version.
- Tests for as much as doesn't require mocking of the SQL driver.

I'm not sure if this is really how we want to do it, so it's WIP for now.

Fixes #93 

## Type of change

## Testing steps

```shell
go test ./...
```


storage/migration.go
```diff
@@ -0,0 +1,129 @@
...
+		tx.Rollback()
+		return err
+	}
+
+	return tx.Commit()
+}
+
+// GetDBVersion reads the current version of the database from the migration info table.
+func GetDBVersion(db *DBStorage) (MigrationVersion, error) {
+	versionRow := db.connection.QueryRow("SELECT version FROM migration_info")
```
tisnik: prob. add something like LIMIT 1, just to inform that just one record is expected?

storage/migration.go
```diff
@@ -0,0 +1,129 @@
...
+
+package storage
+
+import (
+	"database/sql"
+	"fmt"
+)
+
+// MigrationVersion represents a version of the database.
+type MigrationVersion int
```
tisnik: I'd use uint - less conditions in the rest of code

storage/migration.go
```diff
@@ -0,0 +1,129 @@
...
+	var version MigrationVersion = 0
+	err := versionRow.Scan(&version)
+	return version, err
+}
+
+// SetDBVersion attempts to get the database into the specified
+// target version using available migration steps.
+func SetDBVersion(db *DBStorage, targetVer MigrationVersion) error {
+	migrationCount := len(migrations)
+	if targetVer < 0 || targetVer > MigrationVersion(migrationCount) {
```
tisnik: if case of unit (instead of int) code will be shorter a bit :)

storage/migration.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Get current database version.
+	currentVer, err := GetDBVersion(db)
+	if err != nil {
+		return err
+	}
+
+	// Current version is unexpectedly high.
+	if currentVer < 0 || currentVer > MigrationVersion(migrationCount) {
```
tisnik: dtto

storage/migration.go
```diff
@@ -0,0 +1,129 @@
...
+// create the most recent version of the database from scratch.
+var migrations []Migration
+
+// AddMigration adds a new migration to the list of available migrations.
+func AddMigration(m Migration) {
+	migrations = append(migrations, m)
+}
+
+// InitMigrationInfo creates a table containing migration information in the database.
+func InitMigrationInfo(db *DBStorage) error {
```
bond95: I think will be nice to use this function in case if table `migration_info` doesn't exists

storage/migration.go
bond95: Agree, to be 100% sure, it will be better to add LIMIT 1

storage/migration.go
natiiix: It would, perhaps, be easier to `CREATE TABLE IF NOT EXISTS` and run that during initialization of the application. WDYT? The second query would be only executed if the first one changed anything. I believe it's possible to check that.

storage/migration.go
natiiix: Fair enough. I rarely use it because almost all stdlib functions return `int` (and many languages 1. use negative values for errors and 2. don't like implicit conversions between signed and unsigned integers). I'm also not sure how the database driver handles it.

storage/migration.go
natiiix: My bad, I thought the function would somehow cause an error if there were multiple rows. I wanted to use it to make sure there is exactly 1 row in the table.

If, by some error, the table ended up having multiple rows that were not in sync, we could run into mysterious trouble when using `LIMIT 1`.

I will try to figure out some other way to make sure there's exactly one row in the table.

natiiix:
> BTW would be nice to see actual migration scripts. I'm not sure about some more complicated operations like splitting one column into two etc. But it has nothing to do with this code.

That's why the `MigrationStep` is a callback and not just a `string`/`[]string`. Anything that can be done using any number of SQL queries in Go should be doable this way.

storage/migration.go
tisnik: Yup, it might be implemented as a table criteria, if possible (I can imagine one solution which is kinda ugly - have an unique key constraint for another column with constant value).

storage/migration.go
tisnik: Fair enough. Btw IIRC there were proposals to change some stdlib functions to use _proper_ types.

storage/migration.go
tisnik: ok this is now implemented properly. TY

storage/migration.go
natiiix: `MigrationVersion` is now `uint`, so resolved as far as I can tell

natiiix:
> Could you, please, add example of migration and I don't see execution migration in storage init.

@bond95 Could you please see if it's now all ok?

storage/storage.go
```diff
@@ -143,17 +143,32 @@ func getDataSourceFromConfig(configuration Configuration) (string, error) {
...
-		);
-	`)
-	return err
+	if err := InitMigrationInfo(&storage); err != nil {
+		return err
+	}
+
+	ClearMigrations()
+
+	AddMigration(Migration{
```
bond95: I'm not sure about defining migration in `Init` maybe we could move it somewhere else?

natiiix: @bond95 I wanted to put it into `init()` in `migration.go`, but that breaks tests that modify the migration list. Perhaps it shouldn't even be possible to add migrations at runtime ( = they'd be hardcoded in `migration.go` or their own file in the `storage` package and there would be no way to modify the migration list externally at runtime).

bond95: Yes, I think migration should be hardcoded into binary, and may be changed only from point in code.

natiiix: @bond95 So do you want me to remove the `AddMigration` and `ClearMigrations`, including their respective tests?

natiiix: @bond95 So do you want me to remove the `AddMigration` and `ClearMigrations`, including their respective tests?

bond95: @natiiix no, I just want to move it in separate file.

tisnik:
2020/02/26 17:37:17 Making connection to data storage, driver=sqlite3 datasource=./test.db
2020/02/26 17:37:17 table report already exists

does the code tries to create table several times? That's kinda strange

PR #153, opened 2020-02-26
Clusters endpoint rest api tests

tisnik:
# Description

Some more REST API tests.

## Type of change

- REST API tests (no changes in the code)

## Testing steps
Done on CI

### DIFF
--- a/tests/rest/server.go
+++ b/tests/rest/server.go
@@ -16,10 +16,22 @@ limitations under the License.
 
 package tests
 
-import "github.com/verdverm/frisby"
+import (
+	"strconv"
+
+	"github.com/verdverm/frisby"
+)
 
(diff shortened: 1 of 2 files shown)


tests/rest/server.go
```diff
@@ -84,12 +98,55 @@ func checkWrongMethodsForEntryPoint() {
...
+	if err != nil {
+		f.AddError(err.Error())
+	} else {
+		json.Unmarshal(text, &response)
+	}
+	return response
+}
+
+// compare two organization lists
+func compareOrglists(a, b []int) bool {
```
Sergey1011010: maybe it's better to use something like `reflect.DeepEqual`?

tisnik: yup, lemme change that

PR #167, opened 2020-02-27
Basic check for report structure

### LINKED ISSUE #166: Basic check for the content of report consumed from broker
Messages consumed from broker contains report which is now handled as string. It is ok from storage perspective, but we need to check (at least) if all required fields are there.

tisnik:
# Description

Report structure needs to be checked a bit.

Fixes #166

## Type of change

## Testing steps
It is tested by `make test` as usual.

### DIFF
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -53,11 +53,14 @@ type KafkaConsumer struct {
 	client                               sarama.Client
 }
 
+// report represents report send in a message consumed from any broker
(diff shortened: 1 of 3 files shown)


tisnik:
> TYVM!
> 
> Maybe tests can be improved with a test where there are some missing key in the report. Now all the tests are "all or nothing", but anyway it looks good to me

yup homework for me for tomorrow ;)

PR #168, opened 2020-02-27
Refactor storage new

### LINKED ISSUE #165: It might be better to add this check before sql.Open
https://github.com/RedHatInsights/insights-results-aggregator/blob/ab5bb7c9770365641b6e8da7f4aabd715e5dcb1c/storage/storage.go#L110

Now the `default` branch can not be accessed by unit tests, that's one reason ;)

Another reason: the DB connection is keeping opened for unsupported driver

Sergey1011010:
# Description

Fixes #165

## Type of change

## Testing steps

`make before_commit`


storage/storage.go
```diff
@@ -63,8 +62,8 @@ type Storage interface {
 type DBDriver int
 
 const (
-	// DBDriverSQLite shows that db driver is sqlite
-	DBDriverSQLite DBDriver = iota
+	// DBDriverSQLite3 shows that db driver is sqlite
+	DBDriverSQLite3 DBDriver = iota
```
bond95: Instead of using `driverName`, will be better rewrite enum as 
```
	DBDriverSQLite3 DBDriver = "sqlite"
	DBDriverPostgres = "postgres"
```
So the `driverName` variable is not needed anymore.

Sergey1011010: It's not a good idea to compare strings each time we're doing request to the storage

bond95: As I see, we doing check only for one request, is `WriteReportForCluster`, so I don't think it will much slower.

tisnik: just FYI: it will be called at least 30 000x per day (probably with higher frequency later)

Sergey1011010: @bond95 why to make it slower when it's possible to keep it fast?

bond95: Well, ok, then. But my point is, that `driverName` variable is useless, because all needed information we already has in `driverType`, so I suggest to replace it for example with mapper `var DriverMapper = [2]string{"sqlite", "postgres"}`, and when we need name of driver we can get it from array.

Sergey1011010: We need driver's name only in `storage.New()`. So you propose to create an array and access it by index? What if someone changes the order of drivers' constants? 

> But my point is, that driverName variable is useless

We need it here https://github.com/RedHatInsights/insights-results-aggregator/blob/d203c218558f8d112461607240ccfcd13646db25/storage/storage.go#L98 to change the driver name to wrapped one and here https://github.com/RedHatInsights/insights-results-aggregator/blob/d203c218558f8d112461607240ccfcd13646db25/storage/storage.go#L98 to connect to the driver.

bond95: But do we really need to get it from `InitAndGetSQLDriverWithLogs`? Because as I said we already have all needed info, `driverName` is basically duplicate.


> So you propose to create an array and access it by index?

Yes.



> What if someone changes the order of drivers' constants?

Why somebody needs to do that? And even someone will do, then it should be changed in enum as well.

Sergey1011010: > But do we really need to get it from InitAndGetSQLDriverWithLogs

Because we need to create another wrapped driver with different name

https://github.com/RedHatInsights/insights-results-aggregator/blob/d203c218558f8d112461607240ccfcd13646db25/storage/sql_hooks.go#L82

bond95: But driver names are predefined, we can just map driverType to driverName and remove useless return.

Sergey1011010: There are special drivers which are basically wrappers around existing ones for logging

Sergey1011010: I think we can merge this one and you can open an issue for refactoring if you think it would be better

PR #172, opened 2020-02-28
Adding support for errcheck

### LINKED ISSUE #103: Incorporate errorcheck tool into our CI job (linter)
It might be useful to incorporate the following tool into our CI job:
https://github.com/kisielk/errcheck

JoseLSegura:
# Description

errcheck allows to identify possible ignored errors in Go.

This change will execute `errcheck` but won't enforce the result

Fixes #103 

## Type of change

## Testing steps
Use `make errcheck` or `make style` to see the output of errcheck.


Makefile
```diff
@@ -29,7 +29,11 @@ cyclo: ## Run gocyclo
 shellcheck: ## Run shellcheck
 	shellcheck **/*.sh
 
-style: fmt vet lint cyclo shellcheck ## Run all the formatting related commands (fmt, vet, lint, cyclo) + check shell scripts
+errcheck: ## Run errcheck
+	@echo "Running errcheck"
+	-./goerrcheck.sh
```
Bee-lee: What does the dash do?

JoseLSegura: The dash tells `make` to ignore the exit value for that command. So, if `goerrcheck.sh` returns 1 (as in this case while we correct these problems), the "errcheck" rule won't fail

tisnik: Ok basically we have warning right now, ok

JoseLSegura:
> Rights now we ignore return value of errcheck, which, I think, is not okay. There's one problem that this tool shows a lot of not checked errors in defers and I think this is the only error that should be ignored.
> 
> If we unable to configure errcheck, I'd write bash wrapper using grep, which would ignore only lines with defer in the beginning. Maybe even something like `^defer.+close\(\)`

Good point. I will check the errcheck options and try to see if the tool supports ignoring defers.

thank you very much for your review!

Makefile
Bee-lee: I see, thank you

tisnik:
We might want to rerun tests as most issues might be already solved ;)

tisnik:
please rebase, now it should work

PR #204, opened 2020-03-02
Insights rules content

Bee-lee:
# Description

Add the updater part of insights-rules-content

Fixes # (issue)

## Type of change

- Documentation update

## Testing steps


rules/rules.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package rules takes care of fetching and serving the content of Insights
+// rules (descriptions, reasons, etc.) as well as keeping the content
+// up to date by periodically updating them.
+package rules
+
+// RulesConfiguration represents the configuration of rules package
+type RulesConfiguration struct {
```
tisnik: is not it the same as rules/configuration.go?

Bee-lee: I was gonna use https://github.com/robfig/cron which understands the regular cron syntax, I will see how it works, thanks for the article!

Regarding the rules.go, you're right, I left two unsaved files yesterday :)

.gitignore
```diff
@@ -5,3 +5,4 @@ config_my.toml
 coverage.out
 aggregator
 rest-api-tests
+rules/content
```
JoseLSegura: Minor stuff: add the new line

config.toml
```diff
@@ -4,19 +4,23 @@ topic = "ccx.ocp.results"
 group = "aggregator"
 enabled = true
 
-[server]
-address = ":8080"
-api_prefix = "/api/v1/"
-api_spec_file = "openapi.json"
-debug = true
+[logging]
```
JoseLSegura: Is it really necessary to add an empty section? Only asking

JoseLSegura:
I'm wondering why not to use OCP CronJobs https://docs.openshift.com/container-platform/4.3/nodes/jobs/nodes-nodes-jobs.html

For example, we can create a very simple service that downloads the repository, checks if there are new changes in the repository and, if so, copy it to a shared data volume. In the `insights-resultsaggregator` side it only needs to use something that listens to file system changes (something like inotify) to be aware of changes on the repository and then parse and reload the content objects in memory.

WDYT about that approach?

Bee-lee:
After discussing with @quarckster, there is a different and very easy solution.

We discussed the solution in this PR last week when Dima was taking RH exam and was unable to add to the discussion, so hopefully this won't happen in the future, but we won't use anything from this PR.

The solution is roughly:
- add `git clone` to the Dockerfile, so it builds the image with the content.
- add a webhook in the content repository to our Openshift cluster, indicating changes have been made to the repo and the image should be rebuilt
  - if the image rebuild fails for whatever reason, the old image keeps running
  - if the build is successful, the container is seamlessly replaced with zero out-time

PR #205, opened 2020-03-02
Fixed missing configuration sections

### LINKED ISSUE #203: Missing  PROCESSING_ section in configuration.go
https://github.com/RedHatInsights/insights-results-aggregator/blob/master/configuration.go#L101

quarckster:
# Description

Configuration parser missed some sections from config.toml. This PR fixes it.

Fixes #203 

## Type of change

## Testing steps
Tested locally.

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -21,14 +21,15 @@ import (
 	"bufio"
 	"encoding/csv"
 	"fmt"
-	"github.com/deckarep/golang-set"
-	"github.com/spf13/viper"
 	"io"
 	"log"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	mapset "github.com/deckarep/golang-set"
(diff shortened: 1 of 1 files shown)


Sergey1011010:
yes, I'm rewriting it in the way so that we need to modify only one structure when we want to add a new section or something

configuration.go
```diff
@@ -98,6 +105,15 @@ func loadConfiguration(defaultConfigFile string) {
 
 	serverCfg.AutomaticEnv()
 	serverCfg.SetEnvPrefix(envPrefix + "SERVER_")
+
+	processingCfg.AutomaticEnv()
+	processingCfg.SetEnvPrefix(envPrefix + "PROCESSING_")
```
bond95: Could you, please, add this variables to README?

Issue #157, opened 2020-02-27
Processing 'content' for insights rules


tisnik:
As developer I need to be able to process the whole repository with 'content' for insights rules so I'll be able to create in-memory representation of such data to be available via REST API.

Repository with 'content' (templates for messages) for insights rules has specified structure. We need to be able to go through cloned repository and create object from data that will be stored in memory (like a cache). In case of any error, the old version of the object will be used instead so the service won't be downgraded.

See also #156

JoseLSegura:
Will it be more than one repository?

PR #206, opened 2020-03-02
Add core logic for parsing rule content

natiiix:
# Description

Core of the rule content parsing logic, which will be used to get user-friendly description of failing rules.

Solves #157 

## Type of change

## Testing steps

:question:

### DIFF
--- a/content/content.go
+++ b/content/content.go
@@ -42,6 +42,19 @@ type RuleContent struct {
 // RuleContentDirectory contains content for all available rules in a directory.
 type RuleContentDirectory map[string]RuleContent
 
+// readFilesIntoByteArrayPointers reads the contents of the specified files
+// in the base directory and saves them via the specified byte slice pointers.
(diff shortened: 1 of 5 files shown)


content/content.go
```diff
@@ -0,0 +1,134 @@
...
+
+// RuleErrorContent wraps content of a single error key.
+type RuleErrorContent struct {
+	Generic  []byte
+	Metadata []byte
+}
+
+// RuleContent wraps all the content available for a rule into a single structure.
+type RuleContent struct {
+	Summary    []byte
```
JoseLSegura: Go-newbie question: why are you using []byte instead strings on this struct? At some point we will be using them as string, so why not to store them as strings from the beginning?

natiiix: @JoseLSegura Will we ever really be using them as strings? The reason was that the conversion from `[]byte` to `string` can fail and I'm not exactly sure what happens when it does. Plus, it seemed like a waste of performance if we didn't really handle them as strings, which we might not.

PR #209, opened 2020-03-03
Explictely ignoring some error checks

JoseLSegura:
# Description

Fixing some minor errcheck warnings, only ignoring the error value returned by some function calls

## Type of change

## Testing steps
Run the errcheck tool and see only `defer` calls are listed


JoseLSegura:
We are already ignoring them. In addition, it is "only" test code, so I guess we can assume that for the moment

Sergey1011010:
Yes, but we shouldn't ignore them. What about to wrap them to function like `failOnError(t *testing.T, err error)`?

PR #236, opened 2020-03-09
Content updater config

JoseLSegura:
# Description

An small commit adding the possibility of configure the contents directory from the config file or environment variable. Maybe useful for testing in local deployment

## Type of change

## Testing steps
No new tests


JoseLSegura:
Obviously it will need more work (tests), but as first approach to configuration for the content updater and an small test for me to work and understand `viper`

configuration.go
```diff
@@ -140,6 +142,10 @@ func getServerConfiguration() server.Configuration {
 	return config.Server
 }
 
+func getContentPathConfiguration() string {
```
bond95: I don't see any usage of this function, so, please, add example of usage, otherwise I don't see any reason why to have it all if we don't use it.

JoseLSegura: In a short term when the entire content updater feature is integrated, we will need a way to configure where the contents are in order to read and parse them. It will be used at program starts and the result of the call to this function will be passed to the Ivo's package in order to read the rules content repository.

It is truth that right now it is not used, but I think that the Ivo's PR will definitively make use of it.

Thanks for the review and good catch ;-)

JoseLSegura:
> LGTM, are you going to merge it before Ivo's PR?

Right now I don't have enough reviews yet, but it can wait until the Ivo's one is ready

configuration.go
JoseLSegura: Right now it is needed to get the configuration and reload the content because the rest of functionality is WIP, but I think that this will be needed and it could be useful during the development

bond95: After having talk with @natiiix I'm not sure that this feature will be used.

natiiix: I definitely believe we should be able to configure the content path, but it should be enough to have in the `Configuration`.

JoseLSegura: Do you mean to put the `string` path into the `config struct` instead of the whole new struct in `content/configuration.go`?

natiiix: @JoseLSegura I forgot the configuration was split like that, so I guess it does really make sense. It just feels strange to have a single property in the content configuration struct.

JoseLSegura: Yeah, it was strange for me too when writing the file, but I though in that moment that maybe we will have any other configuration property. At the time, apparently we won't need any other "content configuration" stuff, so it can be moved without any problem. I will do it.

Thanks for the review!

JoseLSegura: I updated it. I mimicked the solution provided for `OrgWhiteList`. I tried to find a way to get directly the content -> path from the configuration using tags, but I didn't found a way to do that, so I preferred to keep the same solution that was adopted for `OrgWhiteList` config.

Thanks in advance for your review

bond95:
@JoseLSegura please, fix CI

PR #240, opened 2020-03-09
Added config for devel environment

quarckster:
# Description
Added config for devel environment and updated respectful README section.

## Type of change

- Documentation update


org_whitelist.csv
```diff
@@ -2,3 +2,4 @@ OrgID
 1
 2
 3
+11789772
```
quarckster: This is org id of insight-qe account.

PR #247, opened 2020-03-09
Add second migration for content tables

natiiix:
# Description

Added migration with tables for rule content.

Contributes towards #157 

## Type of change

## Testing steps

Migrations should pass successfully, which by proxy means that unit tests (one of which runs the migrations) should pass.

### DIFF
--- a/migration/migration2.go
+++ b/migration/migration2.go
@@ -1,0 +1,62 @@
+/*
+Copyright © 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migration
+
+import (
+	"database/sql"
+)
+
+var mig2 = Migration{
+	StepUp: func(tx *sql.Tx) error {
+		_, err := tx.Exec(`
+			CREATE TABLE rule (
+				module          VARCHAR PRIMARY KEY,
+				name            VARCHAR NOT NULL,
+				summary         VARCHAR NOT NULL,
+				reason          VARCHAR NOT NULL,
+				resolution      VARCHAR NOT NULL,
+				more_info       VARCHAR NOT NULL
+			)`)
+		if err != nil {
+			return err
+		}
+
+		_, err = tx.Exec(`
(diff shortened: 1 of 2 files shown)


migration/migration2.go
```diff
@@ -0,0 +1,62 @@
...
+		_, err = tx.Exec(`
+			CREATE TABLE rule_error_key (
+				error_key       VARCHAR NOT NULL,
+				rule_module     VARCHAR NOT NULL REFERENCES rule(module),
+				condition       VARCHAR NOT NULL,
+				description     VARCHAR NOT NULL,
+				impact          INTEGER NOT NULL,
+				likelihood      INTEGER NOT NULL,
+				publish_date    TIMESTAMP NOT NULL,
+				status          BOOLEAN NOT NULL,
```
Sergey1011010: maybe it's better to change to isSomething?

migration/migration2.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"database/sql"
+)
+
+var mig2 = Migration{
+	StepUp: func(tx *sql.Tx) error {
+		_, err := tx.Exec(`
+			CREATE TABLE rule (
+				module          VARCHAR PRIMARY KEY,
```
tisnik: module is fqdn right? TPB their repository is a bit inconsistent in this case :)

migration/migration2.go
tisnik: dunno if status is just a boolean

Bee-lee: they have status as "active" or "inactive", no idea why they did this
https://gitlab.cee.redhat.com/insights-rules/insights-content/commit/3f18e52dfcb1cbdca48c8275b1d425b67a99ac83
https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp/blob/master/content/external/rules/cluster_wide_proxy_auth_check/AUTH_OPERATOR_PROXY_ERROR/metadata.yaml

migration/migration2.go
natiiix: I used the names from the YAML files.

natiiix: Also not sure about the `status`, but it was on the whiteboard, so I wrote it down that way.

migration/migration2.go
natiiix: So perhaps `isActive` would be better?

JoseLSegura: The problem is that if `status` can have several values different from "active" or "inactive", it will imply to modify our table, adding more columns to it.

Maybe we can ask some people of the rules teams and check which values are valid as `status` before going ahead

natiiix: @JoseLSegura Just checked with the rule/content team and they said that it should indeed be a boolean value.

migration/migration2.go
tisnik: ok agreed

migration/migration2.go
JoseLSegura: Perfect then. It's OK for me

PR #255, opened 2020-03-10
Load content into DB during initialization

### LINKED ISSUE #157: Processing 'content' for insights rules

natiiix:
# Description

We need content and rule metadata in the database. Therefore we must parse it and load it into the DB before doing anything else.

### TODO

- [x] Add/Modify Unit Tests

Fixes #157 

## Type of change

## Testing steps

Aggregator should start successfully. This requires the `/rules-content/` directory to exist.


aggregator.go
```diff
@@ -142,6 +174,12 @@ func startService() int {
 	var waitGroup sync.WaitGroup
 	exitCode := 0
 
+	prepDbExitCode := prepareDB()
+	if prepDbExitCode != 0 {
+		fmt.Printf(consumerExitedErrorMessage, prepDbExitCode)
```
bond95: Please, use `fmt.Errorf` to be consistent

migration/migration2.go
```diff
@@ -25,7 +25,7 @@ var mig2 = Migration{
 		_, err := tx.Exec(`
 			CREATE TABLE rule (
 				module          VARCHAR PRIMARY KEY,
-				name            VARCHAR NOT NULL,
+				"name"            VARCHAR NOT NULL,
```
bond95: Is this necessary to have here quotes? If yes, maybe will be better to add quotes to other filed names as well? WDYT?

tisnik: imho in this case it is indeed needed

JoseLSegura: Why is it needed? If it is, I agree with @bond95 to put quotes to every column name

natiiix: @bond If none of the others are, `name` is a reserved keyword in PostgreSQL. Some of the others may be too, but probably not in this context.

natiiix: @JoseLSegura We should be probably doing that always everywhere anyways. It's just a bit of a nuisance.

natiiix: I was getting strange errors (`no column "name" in relation "rule"` or something like that) before I added the quotes.

[The documentation is here.](https://www.postgresql.org/docs/8.1/sql-keywords-appendix.html)

JoseLSegura: It's ok for me then. TY

PR #256, opened 2020-03-11
Fixed bug #232

### LINKED ISSUE #232: Repeated string 'GET' that can be written as constant found in server/server.go
other occurrence(s) of 'GET' found in:
```server/server.go:171:97
server/server.go:172:131
server/server.go:173:116
server/server.go:176:79
server/server.go:179:119```

tisnik:
# Description

Fixes #232

## Type of change

## Testing steps
N/A

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -51,9 +51,6 @@ const (
 	defaultConfigFilename = "config"
 
 	consumerExitedErrorMessage = "consumer exited with error code %v"
-
-	// ruleContentDirPath is a path of the directory with the rule contents.
-	ruleContentDirPath = "/rules-content/"
 )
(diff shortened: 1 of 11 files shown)


tisnik:
> Maybe it's better to use `http.MethodGET` constant?

it is :) nice catch

tisnik:
> LGTM, I guess, but this seems **really excessive** and unnecessary.
> 
> Someone should add an exception for strings like these. At some point they'll require empty strings to be stored as a constant too and we'll end up with .NET's `string.Empty` instead of just `""`.

I agreed, OTOH in order to set up a `goconst` we'd need to fix all those minor things

PR #280, opened 2020-03-11
Serve rules content

### LINKED ISSUE #158: Make the 'content' for rules available via REST API
As front end developer I need to be able to have 'content' for rules (messages markdown) available via REST API so I'll be able to process and show messages on web UI.

For the initial version: one rule per REST API call, but we might end with different solution (like mixing results with 'content' in one REST API call).

See also #156 and #157

Bee-lee:
# Description
Processing the cluster report and getting the rules' content from DB. 

- Serhii wrote most unit tests for storage part. Thank you @Sergey1011010 !. I'll add more coverage for server part later
- Deployment server variables updated https://github.com/RedHatInsights/e2e-deploy/pull/1215
- OpenAPI specs file updated

Possible imporvements:
- process report in consumer part so the server doesn't have to unmarshall it every time and storage could possibly just `JOIN` to rules.

In this PR there were also updates to the config files:
- updated all config files with correct `/rules-content` path and changed ports to Kafka to match the ports defined in the minimal stack `docker-compose` file by Dmitrii

Fixes #158 

## Type of change

- Documentation update

## Testing steps
`make before_commit`, tested manually ENV variable for `INSIGHTS_RESULTS_AGGREGATOR__CONTENT__PATH`


Bee-lee:
@JoseLSegura sure, thanks! this is in a really WIP state :)

storage/storage.go
```diff
@@ -266,6 +267,68 @@ func (storage DBStorage) ReadReportForCluster(
...
+	whereInStatement := constructWhereClause(reportRules)
+
+	q := `SELECT rek.error_key, rek.rule_module, rek.description, rek.publish_date,
+		rek.impact, rek.likelihood
+		FROM rule_error_key rek LEFT JOIN rule r ON rek.rule_module = r.module
+		WHERE (rek.error_key, rek.rule_module) IN (%v)`
+
+	q = fmt.Sprintf(q, whereInStatement)
+
+	rows, err := storage.connection.Query(q)
```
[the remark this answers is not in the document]
JoseLSegura: I think you probably already fixed it, but I guess that the problem is a missing `AS` between the table name and its alias

`FROM rule_error_key as rek`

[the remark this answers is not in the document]
Bee-lee: @JoseLSegura That's valid syntax. AS is optional. The query works if I insert some `IN (VALUES ("errorkey", "module"))` but not if I let `.Query(q, ...arguments)` do it.

Bee-lee:
@tisnik Please do already, I just wanted to add unit tests, finalize the response format with UI guys and possibly merge this

server/server.go
```diff
@@ -122,6 +123,40 @@ func (server *HTTPServer) listOfClustersForOrganization(writer http.ResponseWrit
 	}
 }
 
+func getTotalRuleCount(reportRules types.ReportRules) int {
```
tisnik: strictly speaking it should be `uint`, but people don't like unsigned numbers :)

server/server.go
```diff
@@ -141,9 +176,27 @@ func (server *HTTPServer) readReportForCluster(writer http.ResponseWriter, reque
 	} else if err != nil {
 		log.Println("Unable to read report for cluster", err)
 		responses.SendInternalServerError(writer, err.Error())
+	}
+
+	rulesContent, rulesCount, err := server.getContentForRules(writer, report)
+	if err != nil {
+		return
```
tisnik: ah I see, the response is handled in previous method. It is a bit unusual, prob. worth to move the `Send*` here?

Bee-lee: I was considering it, but I wanted to use the same "style" of responses like it's used here (in this function too), 20L before is
```
	organizationID, err := readOrganizationID(writer, request)
	if err != nil {
		// everything has been handled already
		return
	}

	clusterName, err := readClusterName(writer, request)
	if err != nil {
		// everything has been handled already
		return
	}
```
I added the comment for completeness

tisnik: ok, TY

storage/storage_rules_test.go
```diff
@@ -0,0 +1,379 @@
...
+						PublishDate: "1970-01-03 00:00:00",
+						Status:      "active",
+					},
+				},
+			},
+		},
+	}
+)
+
+func TestLoadRuleContentActiveOK(t *testing.T) {
```
JoseLSegura: For the whole file, it could be nice if you add in this commit or lately docstring comments to each function

config.toml
```diff
@@ -25,4 +25,4 @@ pg_params = ""
 log_sql_queries = true
 
 [content]
-path = "/rules_content"
```
quarckster: Please add the same into confi-devel.toml as well. Besides  environment variables should be updated here https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/ccx-data-pipeline/insights-results-aggregator.yml#L72

utils/produce_insights_results/produce.sh
```diff
@@ -19,7 +19,7 @@ do
     for file in *.json
     do
         echo $file
-        kafkacat -b localhost:9092 -P -t ccx.ocp.results $file
+        kafkacat -b localhost:29092 -P -t ccx.ocp.results $file
```
bond95: If we gonna use it with our `docker-compose`, then I think we should update configuration of consumer as well. I mean change `address = "localhost:9092"` -> `address = "localhost:29092"`

Bee-lee: @bond95 @quarckster I was just about to raise the question here if I should remove this or update atleast the `config-devel.toml`. Or even in other configs since in other envs it's set by `KAFKA_BOOTSTRAP_PORT`, right? I think everyone runs the `docker-compose` now but I'm not sure about other configs

quarckster: We should update all config for consistensy sake.

bond95: Agree

Bee-lee: Done

config.toml
Bee-lee: Updated the configs file and created a PR in e2e-deploy
https://github.com/RedHatInsights/e2e-deploy/pull/1215

utils/produce_insights_results/produce.sh
natiiix: Thanks for this! It's quite important that the parts integrate well together. Finding various misconfigured ports is a nightmare.

PR #283, opened 2020-03-11
Add .EditorConfig

### LINKED ISSUE #281: We could add .EditorConfig for consistency
To avoid any further "missing final newline" problems.

natiiix:
# Description

Added EditorConfig to make code style (outside of Go code) slightly more consistent. Same thing as what we did in the external data pipeline a couple of months back.

I'm doing this mainly for YAML config files and Markdown files, which tend to be very inconsistent with their indentation. Also to prevent missing final newlines in files.

If this makes your IDE reformat any file in a strange way, please tweak the .EditorConfig or let me know.

Fixes #281 

## Type of change

## Testing steps

Letting your IDE format any of the existing files shouldn't break it. If it does, let me know.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.go]
+indent_style = tab
+indent_size = unset
+
+[*.{yml,yaml}]
+indent_size = 2
+
+[openapi.json]
+indent_size = 2
+
+[Makefile]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,22 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
```
tisnik: just FYI (not applicable there) - trailing whitespaces are sometimes necessary in AsciiDoc files (because it's grammar is pretty bad in many corner cases).

JoseLSegura:
Nice! Good to have.

.editorconfig
natiiix: @tisnik Also in `.vimrc` and various scripts, where strings are written outside of quotes and you want to write a string ending with whitespace.

In general, however, I assume you don't need trailing whitespace and I've seen a lot of it in the Markdown files.

Issue #331, opened 2020-03-19
CORS headers


bond95:
Need to be added CORS headers for testing frontend on local aggregator

bond95:
I made a few changes to `uhc-portal` to make it work with aggregator without CORS, so basically this issue is not actual.

PR #332, opened 2020-03-19
Added CORS headers

bond95:
# Description

Added CORS headers and mechanism to turn on and turn off it. Also added HTTPS connection

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/331

## Type of change

- Documentation update

## Testing steps

Run any request and check headers.


config-devel.toml
```diff
@@ -15,6 +15,8 @@ address = ":8080"
 api_prefix = "/api/insights-results-aggregator/v1/"
 api_spec_file = "openapi.json"
 debug = true
+use_https=false
+enable_cors=true
```
tisnik: just please add spaces around `=`

tisnik:
Please rebase ;)

server/server.go
```diff
@@ -313,6 +325,15 @@ func (server *HTTPServer) Initialize(address string) http.Handler {
 
 	apiPrefix := server.Config.APIPrefix
 
+	if server.Config.EnableCORS {
+		router.Use(server.addDefaultHeaders)
```
Sergey1011010: maybe it's better to call it addCORSHeaders?

server/server_test.go
```diff
@@ -70,6 +72,12 @@ func checkResponseCode(t *testing.T, expected, actual int) {
 	}
 }
 
+func checkHeaders(t *testing.T, expected, actual string) {
+	if expected != actual {
```
Sergey1011010: Why not to use assert.Equal instead of this function?

bond95: Because it's basically copy of `checkResponseCode`, and in this case there is not so much differences, because headers is one line strings, so it doesn't matter if we check it through `!=` or through `assert`

server/server.go
bond95: Yep, it sound better

server/server_test.go
Sergey1011010: assert shows you old version, new and the diff of objects you're comparing

bond95:
I made a few changes to `uhc-portal` to make it work with aggregator without CORS, so basically this PR is not needed anymore.

bond95:
After discussion with @quarckster I reopen this PR.

tisnik:
please rebase

README.md
```diff
@@ -142,6 +144,8 @@ auth_type = "xrh"
 * `debug` is developer mode that enables some special API endpoints not used on production
 * `auth` turns on or turns authentication
 * `auth_type` set type of auth, it means which header to use for auth `x-rh-identity` or `Authorization`. Can be used only with `auth = true`. Possible options: `jwt`, `xrh`
+* `use_https` is turns on TLS server
```
tisnik: is option to turn on TLS server

README.md
```diff
@@ -142,6 +144,8 @@ auth_type = "xrh"
...
+* `enable_cors` is turns on CORS header, that allows to connect from different hosts (**don't use it in production**)
```
tisnik: is option to turn on ...

server/server.go
```diff
@@ -408,8 +437,13 @@ func (server *HTTPServer) Start() error {
...
+	var err error
 
-	err := server.Serv.ListenAndServe()
+	if server.Config.UseHTTPS {
+		err = server.Serv.ListenAndServeTLS("server.crt", "server.key")
+	} else {
+		err = server.Serv.ListenAndServe()
+	}
 	if err != nil && err != http.ErrServerClosed {
 		log.Error().Err(err).Msg("Unable to start HTTP server")
```
tisnik: actually it can be HTTPS server in this case :)

server/server.go
```diff
@@ -353,6 +353,30 @@ func (server HTTPServer) serveAPISpecFile(writer http.ResponseWriter, request *h
 	http.ServeFile(writer, request, absPath)
 }
 
+// addCORSHeaders - middleware for adding headers that should be in any response
+func (server *HTTPServer) addCORSHeaders(nextHandler http.Handler) http.Handler {
+	return http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Access-Control-Allow-Origin", "*")
+			w.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
```
tisnik: is it added for all endpoints? including ones that are only GET or DELETE, for example?

bond95: Yes

tisnik: why?

bond95: Because it doesn't make any impact to other routes, but if we want to use it only for some single routes that has `OPTIONS` method, then it means we need to make some wrappers, which are I think make whole logic slightly more complicated.

tisnik: ok get it, let's leave it in current state.

tisnik:
please rebase - there were changes required by REST API tests

server/server.go
```diff
@@ -387,10 +416,10 @@ func (server *HTTPServer) Initialize(address string) http.Handler {
...
 	router.HandleFunc(apiPrefix+MainEndpoint, server.mainEndpoint).Methods(http.MethodGet)
-	router.HandleFunc(apiPrefix+ReportEndpoint, server.readReportForCluster).Methods(http.MethodGet)
-	router.HandleFunc(apiPrefix+LikeRuleEndpoint, server.likeRule).Methods(http.MethodPut)
-	router.HandleFunc(apiPrefix+DislikeRuleEndpoint, server.dislikeRule).Methods(http.MethodPut)
-	router.HandleFunc(apiPrefix+ResetVoteOnRuleEndpoint, server.resetVoteOnRule).Methods(http.MethodPut)
+	router.HandleFunc(apiPrefix+ReportEndpoint, server.readReportForCluster).Methods(http.MethodGet, http.MethodOptions)
+	router.HandleFunc(apiPrefix+LikeRuleEndpoint, server.likeRule).Methods(http.MethodPut, http.MethodOptions)
+	router.HandleFunc(apiPrefix+DislikeRuleEndpoint, server.dislikeRule).Methods(http.MethodPut, http.MethodOptions)
+	router.HandleFunc(apiPrefix+ResetVoteOnRuleEndpoint, server.resetVoteOnRule).Methods(http.MethodPut, http.MethodOptions)
 	router.HandleFunc(apiPrefix+ClustersForOrganizationEndpoint, server.listOfClustersForOrganization).Methods(http.MethodGet)
```
Sergey1011010: why no options here?

server/server_test.go
```diff
@@ -58,6 +60,30 @@ func TestMakeURLToEndpoint(t *testing.T) {
 	)
 }
 
+func TestAddCORSHeaders(t *testing.T) {
+	mockStorage := helpers.MustGetMockStorage(t, true)
+	defer helpers.MustCloseStorage(t, mockStorage)
+
+	err := mockStorage.WriteReportForCluster(testdata.OrgID, testdata.ClusterName, "{}", time.Now())
+	if err != nil {
```
Sergey1011010: Would be slightly better to use `helpers.FailOnError(t, err)`

server/server.go
bond95: Yes, `OPTIONS` is only for routes that used in UI.

server/server_test.go
bond95: Yes, could be.

Issue #333, opened 2020-03-19
Make local auth between OCM UI and aggregator


bond95:
For local testing purposes there should be establish local authorization as well. Because of aggregator takes data directly from OCM, there should be JWT token parsed directly.

PR #334, opened 2020-03-19
Establish auth

bond95:
# Description

Now OCM UI can be auth directly with aggregator.
This PR is based on https://github.com/RedHatInsights/insights-results-aggregator/pull/332 , so shouldn't be merged before it. 

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/333

## Type of change

## Testing steps

Allow `auth` and `debug` in configuration and try to make request with JWT token.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -20,10 +20,11 @@ package server
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	jwt "github.com/dgrijalva/jwt-go"
 	"net/http"
+	"strings"
 
 	"github.com/RedHatInsights/insights-results-aggregator/types"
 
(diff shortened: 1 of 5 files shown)


server/auth.go
```diff
@@ -65,22 +91,36 @@ func (server *HTTPServer) Authentication(next http.Handler) http.Handler {
 
 		decoded, err := base64.StdEncoding.DecodeString(tokenHeader) //Decode token to JSON string
 		if err != nil {                                              //Malformed token, returns with http code 403 as usual
+			fmt.Println(err)
```
tisnik: these won't be logged properly. pls. use Log.Error().Err(err) instead

server/auth.go
```diff
@@ -53,10 +54,35 @@ type Token struct {
...
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		tokenHeader := r.Header.Get("x-rh-identity") //Grab the token from the header
+		// Skip authentication for OPTIONS method, because Chrome don't send Authorization header with it
+		if r.Method == "OPTIONS" {
+			next.ServeHTTP(w, r)
+		}
+		var tokenHeader string
+		// In case of testing on local machine we don't take x-rh-identity header, but instead Authorization with JWT token in it
+		if server.Config.Debug {
+			tokenHeader = r.Header.Get("Authorization") //Grab the token from the header
```
tisnik: might be splitted into own function later

PR #361, opened 2020-03-23
Basic reports REST API tests

tisnik:
# Description

Very basic (ATM) REST API tests for `results` endpoint.

## Type of change

- Tests (no changes in the code)
- Documentation update

## Testing steps

You can run `make rest_api_tests` as usual.

### DIFF
--- a/tests/rest/server.go
+++ b/tests/rest/server.go
@@ -312,6 +312,6 @@ func ServerTests() {
 	checkReportEndpointForImproperOrganization()
 	checkReportEndpointWrongMethods()
 
-	// tests for OpenAPI specification that is accessibla via its endpoint as well
+	// tests for OpenAPI specification that is accessible via its endpoint as well
 	checkOpenAPISpecifications()
 }
--- a/tests/rest/server.go
+++ b/tests/rest/server.go
@@ -14,22 +14,21 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/* REST API tests for following endpoints:
-
-apiPrefix
-
-apiPrefix+"organizations"
-
-apiPrefix+"report/{organization}/{cluster}"
-
-apiPrefix+"clusters/{cluster}/rules/{rule_id}/like"
-
-apiPrefix+"clusters/{cluster}/rules/{rule_id}/dislike"
-
-apiPrefix+"clusters/{cluster}/rules/{rule_id}/reset_vote"
-
-apiPrefix+"organizations/{organization}/clusters"
-*/
+// Package tests contains REST API tests for following endpoints:
+//
+// apiPrefix
+//
+// apiPrefix+"organizations"
+//
+// apiPrefix+"report/{organization}/{cluster}"
+//
+// apiPrefix+"clusters/{cluster}/rules/{rule_id}/like"
+//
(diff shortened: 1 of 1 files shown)


tests/rest/server.go
```diff
@@ -210,15 +288,30 @@ func checkOpenAPISpecifications() {
...
+
+	// tests for REST API endpoints apiPrefix+"report/{organization}/{cluster}"
+	checkReportEndpointForKnownOrganizationAndKnownCluster()
+	checkReportEndpointForKnownOrganizationAndUnknownCluster()
+	checkReportEndpointForUnknownOrganizationAndKnownCluster()
+	checkReportEndpointForUnknownOrganizationAndUnknownCluster()
+	checkReportEndpointForImproperOrganization()
+	checkReportEndpointWrongMethods()
+
+	// tests for OpenAPI specification that is accessibla via its endpoint as well
```
quarckster: `// tests for OpenAPI specification that is accessible via its endpoint as well`

tests/rest/server.go
```diff
@@ -199,6 +217,66 @@ func checkClustersEndpointWrongMethods() {
 	}
 }
 
+func constructURLForResultForOrgCluster(organizationID string, clusterID string) string {
```
Sergey1011010: btw, I'm implementing something similar in my branch for covering server with tests

```> cat server/endpoints.go
package server

import (
        "fmt"
        "regexp"
)

const (
        MainEndpoint                    = ""
        DeleteOrganizationsEndpoint     = "organizations/{organizations}"
        DeleteClustersEndpoint          = "clusters/{clusters}"
        OrganizationsEndpoint           = "organizations"
        ReportEndpoint                  = "report/{organization}/{cluster}"
        LikeRuleEndpoint                = "clusters/{cluster}/rules/{rule_id}/like"
        DislikeRuleEndpoint             = "clusters/{cluster}/rules/{rule_id}/dislike"
        ResetVoteOnRuleEndpoint         = "clusters/{cluster}/rules/{rule_id}/reset_vote"
        ClustersForOrganizationEndpoint = "organizations/{organization}/clusters"
        MetricsEndpoint                 = "metrics"
)

// MakeURLToEndpoint creates URL to endpoint, use constants from file endpoints.go
func MakeURLToEndpoint(apiPrefix, endpoint string, args ...interface{}) string {
        re := regexp.MustCompile(`\{[a-zA-Z_0-9]+\}`)
        endpoint = re.ReplaceAllString(endpoint, "%v")
        return apiPrefix + fmt.Sprintf(endpoint, args...)
}
```

so that we don't have to change endpoint it unit tests and rest api tests if we decide to change url

tests/rest/server.go
tisnik: where's problem?

tests/rest/server.go
tisnik: yup for future PR

tests/rest/server.go
quarckster: accessibla should be accessible

tisnik: ah, my eyes! ok sorry

tisnik: fixed

PR #366, opened 2020-03-23
Producer tests

### LINKED ISSUE #365: Cover producer with tests

JoseLSegura:
# Description

A little refactor to the producer package and adding unit tests

Fixes #365 

## Type of change

## Testing steps
CI

### DIFF
--- a/broker/configuration.go
+++ b/broker/configuration.go
@@ -24,6 +24,7 @@ import (
 type Configuration struct {
 	Address      string     `mapstructure:"address" toml:"address"`
 	Topic        string     `mapstructure:"topic" toml:"topic"`
+	PublishTopic string     `mapstructure:"publish_topic" toml:"publish_topic"`
 	Group        string     `mapstructure:"group" toml:"group"`
 	Enabled      bool       `mapstructure:"enabled" toml:"enabled"`
 	OrgWhitelist mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
--- a/producer/producer.go
+++ b/producer/producer.go
@@ -57,10 +57,10 @@ func New(brokerCfg broker.Configuration) (*KafkaProducer, error) {
 // ProduceMessage produces message to selected topic. That function returns
 // partition ID and offset of new message or an error value in case of any
 // problem on broker side.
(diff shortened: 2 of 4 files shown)


broker/configuration.go
```diff
@@ -24,6 +24,7 @@ import (
 type Configuration struct {
 	Address      string     `mapstructure:"address" toml:"address"`
 	Topic        string     `mapstructure:"topic" toml:"topic"`
+	PublishTopic string     `mapstructure:"publish_topic" toml:"publish_topic"`
```
Sergey1011010: Why not to use `Topic`?

producer/producer.go
```diff
@@ -28,31 +28,55 @@ import (
...
+	return &KafkaProducer{
+		Configuration: brokerCfg,
+		Producer:      producer,
+	}, nil
+}
+
+// ProduceMessage produces message to selected topic. That function returns
+// partition ID and offset of new message or an error value in case of any
+// problem on broker side.
+func (producer *KafkaProducer) ProduceMessage(message string) (partition int32, offset int64, errout error) {
```
Sergey1011010: why do you use named return value here? It doesn't seem to be necessary.

producer/producer_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	mockProducer := mocks.NewSyncProducer(t, nil)
+	mockProducer.ExpectSendMessageAndSucceed()
+	producer := producer.KafkaProducer{
+		Configuration: brokerCfg,
+		Producer:      mockProducer,
+	}
+
+	_, _, err := producer.ProduceMessage("Hello world")
+	assert.Equal(t, err, nil)
```
Sergey1011010: In most of cases like this we use helpers.FailOnError(t, err) so it's better to follow one standard :)

broker/configuration.go
JoseLSegura: When we need to use the `producer` package probably we want to use a different topic for consuming and producing messages

producer/producer.go
JoseLSegura: Agree

producer/producer_test.go
JoseLSegura: Ok! Thank you!

Issue #328, opened 2020-03-19
Users from different organizations can access to data under org id they don't belong


quarckster:
Step to reproduce:
1. Upload two archives using two different accounts with different org id.
2. Fetch `/organization` using the credentials of these two users

Actual result:
Both users can see all organizations

Expected results:
Only the organization that the user belongs to should be shown

3. Fetch `/organizations/{org_id}/clusters` using the credentials of these two users

Actual result:
Both users can see all cluster even from under org id that they don't belong to

Expected results:
User should have access only to clusters under his org id

tisnik:
Now it is possible to enforce auth. mode for server, but the check for orgID seems to be missing. It is a regression, so I'm reopening this bug.

PR #389, opened 2020-03-24
Fixed permission issue

bond95:
# Description

Added checks by orgId

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/328

## Type of change

## Testing steps

Enable auth and try to get list of clusters for another organisation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DATA-DOG/go-sqlmock v1.4.1
-	github.com/RedHatInsights/insights-operator-utils v0.0.0-20200323131325-73813484635c
+	github.com/RedHatInsights/insights-operator-utils v0.0.0-20200212101518-5819e1fc31a9
 	github.com/Shopify/sarama v1.26.0
 	github.com/bitly/go-simplejson v0.5.0 // indirect
 	github.com/deckarep/golang-set v1.7.1
--- a/go.sum
+++ b/go.sum
@@ -13,8 +13,6 @@ github.com/RedHatInsights/insights-operator-utils v0.0.0-20200128100540-06c122fd
 github.com/RedHatInsights/insights-operator-utils v0.0.0-20200128100540-06c122fd8a86/go.mod h1:c6ReBK57bYPBl3DCb03lo3Jwr+ORT/9XUdlTwzhKQP8=
 github.com/RedHatInsights/insights-operator-utils v0.0.0-20200212101518-5819e1fc31a9 h1:LK+3X6wSHRHzBV/jir+4GqsIG77hYxE5yj5ai7u8RVg=
 github.com/RedHatInsights/insights-operator-utils v0.0.0-20200212101518-5819e1fc31a9/go.mod h1:c6ReBK57bYPBl3DCb03lo3Jwr+ORT/9XUdlTwzhKQP8=
-github.com/RedHatInsights/insights-operator-utils v0.0.0-20200323131325-73813484635c h1:DK7vkE+F84PpDAAilahJHrY1DSgotzmr097ggoiP+q0=
-github.com/RedHatInsights/insights-operator-utils v0.0.0-20200323131325-73813484635c/go.mod h1:c6ReBK57bYPBl3DCb03lo3Jwr+ORT/9XUdlTwzhKQP8=
 github.com/Shopify/sarama v1.13.0/go.mod h1:FVkBWblsNy7DGZRfXLU0O9RCGt5g3g3yEuWXgklEdEo=
 github.com/Shopify/sarama v1.19.0/go.mod h1:FVkBWblsNy7DGZRfXLU0O9RCGt5g3g3yEuWXgklEdEo=
 github.com/Shopify/sarama v1.26.0 h1:C+zFi+/NJdfeJgZWbu+WaLgk4NcsbmqfFTKsoJmR39U=
--- a/server/auth.go
+++ b/server/auth.go
@@ -40,7 +40,7 @@ const (
 
 // Internal contains information about organization ID
 type Internal struct {
-	OrgID string `json:"org_id"`
+	OrgID types.OrgID `json:"org_id,string"`
 }
 
 // Identity contains internal user info
@@ -57,7 +57,7 @@ type Token struct {
 // JWTPayload is structure that contain data from parsed JWT token
 type JWTPayload struct {
 	AccountNumber types.UserID `json:"account_number"`
-	OrgID         string       `json:"org_id"`
+	OrgID         types.OrgID  `json:"org_id,string"`
 }
 
 // Authentication middleware for checking auth rights
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -43,8 +43,9 @@ var configAuth2 = server.Configuration{
 // TestMissingAuthToken checks how the missing auth. token header (expected in HTTP request) is handled
 func TestMissingAuthToken(t *testing.T) {
 	helpers.AssertAPIRequest(t, nil, &configAuth, &helpers.APIRequest{
-		Method:   http.MethodGet,
-		Endpoint: server.OrganizationsEndpoint,
+		Method:       http.MethodGet,
+		Endpoint:     server.ClustersForOrganizationEndpoint,
+		EndpointArgs: []interface{}{1},
 	}, &helpers.APIResponse{
 		StatusCode: http.StatusForbidden,
 		Body:       `{"status": "Missing auth token"}`,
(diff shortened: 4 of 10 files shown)


server/auth.go
```diff
@@ -41,7 +41,7 @@ const (
 
 // Internal contains information about organization ID
 type Internal struct {
-	OrgID string `json:"org_id"`
+	OrgID uint64 `json:"org_id,string"`
```
tisnik: actually it is `int32`, limit set by Postgres

server/router_utils.go
```diff
@@ -141,14 +141,29 @@ func readClusterName(writer http.ResponseWriter, request *http.Request) (types.C
...
+	return types.OrgID(organizationID), err
+}
+
+func checkPermissions(writer http.ResponseWriter, request *http.Request, orgID types.OrgID, auth bool) error {
+	identityContext := request.Context().Value(ContextKeyUser)
+	if identityContext != nil && auth {
+		identity := identityContext.(Identity)
+		if types.OrgID(identity.Internal.OrgID) != orgID {
+			const message = "You have no permissions to get or change info about this organization"
+			fmt.Println(message)
```
tisnik: please use structured logging for this

server/server.go
```diff
@@ -240,6 +240,22 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
 		return
 	}
 
+	if server.Config.Auth {
+		orgID, err := server.Storage.GetOrgIDByClusterID(clusterID)
```
tisnik: ok, let's hope it won't have performance issues

server/auth.go
tisnik: basically positive subset of `int32` to be more precise, ugh

server/auth_test.go
```diff
@@ -42,7 +42,7 @@ var configAuth2 = server.Configuration{
 func TestMissingAuthToken(t *testing.T) {
 	helpers.AssertAPIRequest(t, nil, &configAuth, &helpers.APIRequest{
 		Method:   http.MethodGet,
-		Endpoint: server.OrganizationsEndpoint,
+		Endpoint: configAuth.APIPrefix + "organizations/1/clusters",
```
Sergey1011010: it's better to set api prefix in config, it's gonna be added to url by AssertAPIRequest automatically

server/router_utils.go
```diff
@@ -141,14 +141,29 @@ func readClusterName(writer http.ResponseWriter, request *http.Request) (types.C
...
 // readOrganizationID retrieves organization id from request
 // if it's not possible, it writes http error to the writer and returns error
-func readOrganizationID(writer http.ResponseWriter, request *http.Request) (types.OrgID, error) {
+func readOrganizationID(writer http.ResponseWriter, request *http.Request, auth bool) (types.OrgID, error) {
 	organizationID, err := getRouterPositiveIntParam(request, "organization")
 	if err != nil {
 		handleOrgIDError(writer, err)
 		return 0, err
 	}
+	err = checkPermissions(writer, request, types.OrgID(organizationID), auth)
```
Sergey1011010: maybe it's better to check it somwhere else like in authorization middleware? Because router_utils is supposed to be only for parsing parameters from route.

storage/storage.go
```diff
@@ -253,6 +254,26 @@ func (storage DBStorage) ListOfClustersForOrg(orgID types.OrgID) ([]types.Cluste
 	return clusters, nil
 }
 
+// ListOfClustersForOrg reads list of all clusters fro given organization
+func (storage DBStorage) GetOrgIDByClusterID(cluster types.ClusterName) (types.OrgID, error) {
+	rows, err := storage.connection.Query("SELECT org_id FROM report WHERE cluster = $1 ORDER BY org_id LIMIT 1", cluster)
```
Sergey1011010: Why do you have `order by` here?

Sergey1011010: and also, you could probably use QueryRow

storage/storage_test.go
```diff
@@ -166,6 +166,19 @@ func TestDBStorageReadReportForCluster(t *testing.T) {
...
 }
 
+// TestDBStorageGetOrgIDByClusterID check the behaviour of method GetOrgIDByClusterID
+func TestDBStorageGetOrgIDByClusterID(t *testing.T) {
+	mockStorage := helpers.MustGetMockStorage(t, true)
+	defer helpers.MustCloseStorage(t, mockStorage)
+
+	writeReportForCluster(t, mockStorage, testOrgID, testClusterName, `{"report":{}}`)
+	orgID, err := mockStorage.GetOrgIDByClusterID(testClusterName)
+	if err != nil {
```
Sergey1011010: it's better to use helpers.FailOnError(t, err)

server/auth.go
bond95: but in types we have:
```Go
type OrgID uint64
```

server/auth_test.go
bond95: I don't quite understand what do you mean?

server/router_utils.go
bond95: Actually no, because auth middleware is not processing URL, so it should do route handler

storage/storage.go
bond95: > Why do you have `order by` here?

Basically it's just to be sure that we get first orgID, it can be easily removed, but I prefer to leave it as it is, just to be sure. 


> and also, you could probably use QueryRow

Yes, thank you

server/auth.go
tisnik: it is wrong, at least on storage side, and causes strange errors

tisnik: ie. might be fixed elsewhere

storage/storage.go
tisnik: it is unique according to DB schema

tisnik: will be nice to see execution plans for both alternatives

server/auth_test.go
Sergey1011010: we're already adding apiprefix from server config here https://github.com/RedHatInsights/insights-results-aggregator/blob/master/tests/helpers/http.go#L82

server/router_utils.go
Sergey1011010: it has everything sent in request actually, but ok, let's keep it as is.

server/auth_test.go
bond95: ok, so I'll remove it

tisnik:
please rebase

bond95:
@tisnik @Sergey1011010 please, take a look

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/bitly/go-simplejson v0.5.0 // indirect
 	github.com/deckarep/golang-set v1.7.1
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
+	github.com/droptheplot/abcgo v0.0.0-20171120220436-23529565504c // indirect
```
tisnik: is it needed for the service?

go.sum
```diff
@@ -87,6 +87,8 @@ github.com/deckarep/golang-set v1.7.1/go.mod h1:93vsz/8Wt4joVM7c2AVqh+YRMiUSc14y
 github.com/dgrijalva/jwt-go v3.2.0+incompatible h1:7qlOGliEKZXTDg6OTjfoBKDXWrumCAMpl/TFQ4/5kLM=
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/dgryski/go-sip13 v0.0.0-20181026042036-e10d5fee7954/go.mod h1:vAd38F8PWV+bWy6jNmig1y/TA+kYO4g3RSRF0IAv0no=
+github.com/droptheplot/abcgo v0.0.0-20171120220436-23529565504c h1:I2n/JrkM6aUq2IjnbUvYGtd3cCLqnv7DFt4iWBrM6N0=
+github.com/droptheplot/abcgo v0.0.0-20171120220436-23529565504c/go.mod h1:wEtublsjLnIljeLoXn1uigj9wBvAcilXg18hXmGW3B0=
```
tisnik: dtto

server/auth_test.go
```diff
@@ -76,24 +76,11 @@ func TestJWTTokenMalformed(t *testing.T) {
 	})
 }
 
-// TestJWTToken checks authorization through Authorization header
```
tisnik: was this test incorrect?

storage/storage.go
tisnik: is `order by` and `limit` there because of insufficient DB schema constraints?

storage/storage.go
```diff
@@ -255,6 +256,26 @@ func (storage DBStorage) ListOfClustersForOrg(orgID types.OrgID) ([]types.Cluste
...
+// GetOrgIDByClusterID reads OrgID for specified cluster
+func (storage DBStorage) GetOrgIDByClusterID(cluster types.ClusterName) (types.OrgID, error) {
+	rows, err := storage.connection.Query("SELECT org_id FROM report WHERE cluster = $1 ORDER BY org_id LIMIT 1", cluster)
+	if err != nil {
+		return 0, err
+	}
+	defer closeRows(rows)
+
+	var orgID uint64
+	for rows.Next() {
```
tisnik: `if rows.Next()`

server/auth_test.go
```diff
@@ -44,7 +44,7 @@ var configAuth2 = server.Configuration{
 func TestMissingAuthToken(t *testing.T) {
 	helpers.AssertAPIRequest(t, nil, &configAuth, &helpers.APIRequest{
 		Method:   http.MethodGet,
-		Endpoint: server.OrganizationsEndpoint,
+		Endpoint: "organizations/1/clusters",
```
Sergey1011010: Sorry, forgot to tell in previous review, you can use `server.ClustersForOrganizationEndpoint` and add `EndpointArgs: []interface{}{1}` like here https://github.com/RedHatInsights/insights-results-aggregator/blob/master/server/server_test.go#L121

server/auth.go
```diff
@@ -40,7 +40,7 @@ const (
 
 // Internal contains information about organization ID
 type Internal struct {
-	OrgID string `json:"org_id"`
+	OrgID uint32 `json:"org_id,string"`
```
Sergey1011010: maybe it's better to use `types.OrgID`?

go.mod
bond95: It was updated automatically, so if you want I can remove this files.

tisnik: it was updated by running `abcgo.sh`, but not by running `go build`. I'd vote to checkout these two files.

storage/storage.go
bond95: For rows mostly used `for` to check more then one, but yes, here can be `if` as well

storage/storage.go
bond95: As I wrote before, I use `order by` just to be sure that we get first OrgId, even if by DB schema it will be unique.

server/auth_test.go
bond95: I don't even remember how I removed it, maybe it was while rebasing. Now it's back

Issue #354, opened 2020-03-23
Improper HTTP response code 500 is returned for `results` endpoint when cluster ID has improper format


tisnik:
For cluster name set to `foobar` HTTP code 500 is returned:

```
{"level":"error","error":"invalid UUID length: 6","time":"2020-03-23T12:17:50+01:00","message":"invalid cluster name format: 'foobar'"}
FAIL  [Check the end point to return list of clusters for known organization by HTTP GET method]
        -  Expected Status 200, but got 500: "500 Internal Server Error
```

## Expected value:
Something from group 4xx

## Reproducer:

```
$ curl -v localhost:8080/api/v1/report/1/foobar
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /api/v1/report/1/foobar HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8080
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json; charset=utf-8
< Date: Mon, 23 Mar 2020 11:20:15 GMT
< Content-Length: 51
< 
{"status":"invalid cluster name format: 'foobar'"}
* Connection #0 to host localhost left intact
```

Bee-lee:
closed by #393

Issue #385, opened 2020-03-24
[Fuzz testing] PUT /clusters/000000000000000000000000000000000000/rules/0/like returns 500


quarckster:
Response from the server:
```
{"status":"invalid cluster name format: '000000000000000000000000000000000000'"}
{"status":"Unable to get user id"}
```
This also applicable for:
PUT /clusters/000000000000000000000000000000000000/rules/0/dislike
PUT /clusters/000000000000000000000000000000000000/rules/0/reset_vote

Sergey1011010:
May be fixed by https://github.com/RedHatInsights/insights-results-aggregator/pull/367 I changed some error status codes there to correct ones https://github.com/Sergey1011010/insights-results-aggregator/blob/cover-server-with-tests/server/router_utils.go#L97 and also fixed multiple json issue in some responses

tisnik:
Well yes and no. I'm already having reproducers, these are as important as code fix :)

Issue #384, opened 2020-03-24
[Fuzz testing] GET /report/000000000000000000000000000000000000/1 returns 500


quarckster:
Response from the server:
```
{"status":"Error during parsing param organization with value 000000000000000000000000000000000000. Error: positive value expected"}
```

PR #393, opened 2020-03-24
Improper http response in results 500

Bee-lee:
u# Description

Some bugfixing and refactoring which was mostly overwritten by the Server unit tests PR.
It was supposed to fix issues #354 , at that time unreported problems I noticed #385 and #384   

These issues should be fixed (#354, #385 #384), I'm intentionally not putting Fixes keyword, so that i can confirm it.

## Type of change

## Testing steps
tested endpoints and unit tests

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -170,12 +170,9 @@ func (server *HTTPServer) readReportForCluster(writer http.ResponseWriter, reque
 	}
 
 	report, lastChecked, err := server.Storage.ReadReportForCluster(organizationID, clusterName)
-	if _, ok := err.(*storage.ItemNotFoundError); ok {
-		responses.Send(http.StatusNotFound, writer, err.Error())
-		return
-	} else if err != nil {
+	if err != nil {
 		log.Error().Err(err).Msg("Unable to read report for cluster")
-		responses.SendInternalServerError(writer, err.Error())
+		handleServerError(writer, err)
 		return
 	}
 
@@ -236,15 +233,15 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
 	if err != nil {
 		const message = "Unable to get user id"
 		log.Error().Err(err).Msg(message)
-		responses.Send(http.StatusInternalServerError, writer, message)
+		responses.SendInternalServerError(writer, message)
 		return
 	}
 
 	err = server.Storage.VoteOnRule(clusterID, ruleID, userID, userVote)
 	if err != nil {
-		responses.Send(http.StatusInternalServerError, writer, err.Error())
+		responses.SendInternalServerError(writer, err.Error())
 	} else {
-		responses.Send(http.StatusOK, writer, responses.BuildOkResponse())
+		responses.SendResponse(writer, responses.BuildOkResponse())
 	}
 }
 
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -236,15 +236,15 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
 	if err != nil {
 		const message = "Unable to get user id"
 		log.Error().Err(err).Msg(message)
-		responses.Send(http.StatusInternalServerError, writer, message)
+		responses.SendError(writer, message)
```
Sergey1011010: it seems to be more like a server error, because we wouldn't be authorized if user didn't send auth token or sent the wrong one

server/server.go
```diff
@@ -236,15 +236,15 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
...
 		return
 	}
 
 	err = server.Storage.VoteOnRule(clusterID, ruleID, userID, userVote)
 	if err != nil {
-		responses.Send(http.StatusInternalServerError, writer, err.Error())
+		responses.SendInternalServerError(writer, err.Error())
 	} else {
-		responses.Send(http.StatusOK, writer, responses.BuildOkResponse())
+		responses.SendResponse(writer, responses.BuildOkResponse())
```
Sergey1011010: as for me Send with specifying a status code seems to be more readable, but it's a nitpick :)

server/server.go
Bee-lee: how can this happen then? when the user doesn't supply the credentials, it returns Auth error, before reaching the func, right?

Bee-lee: and if auth were to be disabled, isn't it a User error when he doesn't supply the identity?

server/server.go
Bee-lee: i'm just trying to follow a standard, since I thought we use those utils functions everywhere, it just wasnt nice trying to find all the places where X response is sent. :D

server/server.go
Sergey1011010: only if server put wrong type in authentication middleware https://github.com/RedHatInsights/insights-results-aggregator/blob/c8cec81e336323fef16fe6a86fa51faae6703bc7/server/auth.go#L130

Bee-lee: @Sergey1011010 so wait, can this happen entirely by user fault? Or is it really a combination of server conf that would cause that to happen even if the request is correct?

I mean I'm trying to understand if it this error can be fixed by the user (eg. by sending the correct identity), then it should be 4xx, but if not then i of course agree with you

Sergey1011010: No, the user has nothing to do with it, it can only happen if someone modifies this line of code https://github.com/RedHatInsights/insights-results-aggregator/blob/c8cec81e336323fef16fe6a86fa51faae6703bc7/server/auth.go#L122 and puts different type there

Bee-lee: ok, I changed it back

Bee-lee:
I made some more changes now, please have a look at them, guys:

- the returning errors from insights-operator-utils are now checked (I was just the person to first update -utils in go.mod since the changes)
- error responses are now sent from one place server/errors.go
- only successful responses are sent from server/server.go
- created Auth error to further unify error sending
- logging in auth
- created some issues for further refactoring
    - https://github.com/RedHatInsights/insights-operator-utils/issues/39
    - https://github.com/RedHatInsights/insights-operator-utils/issues/38

Issue #394, opened 2020-03-24
Invalid/Malformed auth token


quarckster:
Steps to reproduce:

```
curl -vvv -Lk https://ci.cloud.redhat.com/api/insights-results-aggregator/v1 -u user:password
{"status":"Invalid/Malformed auth token"}
```
Aggregator should return `{"status":"ok"}`

PR #395, opened 2020-03-24
Added IsLocalInstance option

bond95:
# Description

Replaced `debug` option with `is_local_instance` to define which auth header to use.

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/394

## Type of change

## Testing steps

Just run with `is_local_instance = true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ api_prefix = "/api/v1/"
 api_spec_file = "openapi.json"
 debug = true
 auth = true
+auth_type = "xrh"
 ```
 
 * `address` is host and port which server should listen to
@@ -140,6 +141,7 @@ auth = true
 * `api_spec_file` is the location of a required OpenAPI specifications file
 * `debug` is developer mode that enables some special API endpoints not used on production
 * `auth` turns on or turns authentication
+* `auth_type` set type of auth, it means which header to use for auth `x-rh-identity` or `Authorization`. Can be used only with `auth = true`. Possible options: `jwt`, `xrh`
 
 ## Local setup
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -133,13 +133,15 @@ api_prefix = "/api/v1/"
 api_spec_file = "openapi.json"
 debug = true
 auth = true
+is_local_instance = false
```
Sergey1011010: maybe it's better to name it use_jwt, or use_h_rh_identity?

Sergey1011010: or to make a string `auth_type`

tisnik: +1 that's better option

tisnik: must be `auth_type`

README.md
```diff
@@ -133,13 +133,15 @@ api_prefix = "/api/v1/"
...
 ```
 
 * `address` is host and port which server should listen to
 * `api_prefix` is prefix for RestAPI path
 * `api_spec_file` is the location of a required OpenAPI specifications file
 * `debug` is developer mode that enables some special API endpoints not used on production
 * `auth` turns on or turns authentication
+* `is_local_instance` change auth header from `x-rh-identity` to `Authorization`. Can be used only with `auth = true`
```
tisnik: must be `auth_type`

config.toml
```diff
@@ -16,6 +16,7 @@ api_prefix = "/api/v1/"
 api_spec_file = "openapi.json"
 debug = true
 auth = false
+is_local_instance = false
```
tisnik: must be `auth_type`

README.md
bond95: Oops, missed it, thank you

PR #404, opened 2020-03-26
Validate cluster and rule ids

### LINKED ISSUE #295: Implement validation of cluster_id for like, dislike and rest_vote endpoints
Currently, any authenticated user can vote on any non-existing cluster. We should prevent this behavior. If a user tries to access to non-existsing cluster 404 should be returned.

### LINKED ISSUE #296: Implement validation of rule_id for like, dislike and reset_vote endpoints.
Should be started after finishing #280. The use case is similar to #295.

### LINKED ISSUE #398: Implement a check of foreign keys in testing env
now all foreign key errors are ignored

Sergey1011010:
# Description

Fixes #295
Fixes #296
Fixes #398

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -261,7 +261,7 @@ func (server *HTTPServer) validateClusterAndRuleID(
 ) bool {
 	clusterExists, err := server.Storage.CheckIfClusterExists(clusterID)
 	if err != nil {
-		log.Err(err)
+		log.Error().Err(err)
 		responses.Send(http.StatusInternalServerError, writer, err.Error())
 		return false
 	}
@@ -276,7 +276,7 @@ func (server *HTTPServer) validateClusterAndRuleID(
 
 	ruleExists, err := server.Storage.CheckIfRuleExists(ruleID)
 	if err != nil {
-		log.Err(err)
+		log.Error().Err(err)
 		responses.Send(http.StatusInternalServerError, writer, err.Error())
 		return false
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -243,33 +243,8 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
 		return
 	}
 
-	clusterExists, err := server.Storage.CheckIfClusterExists(clusterID)
-	if err != nil {
-		log.Err(err)
-		responses.Send(http.StatusInternalServerError, writer, err.Error())
-		return
-	}
-	if !clusterExists {
-		responses.Send(
-			http.StatusNotFound,
-			writer,
-			(&storage.ItemNotFoundError{ItemID: clusterID}).Error(),
-		)
-		return
-	}
-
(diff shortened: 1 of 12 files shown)


server/server.go
```diff
@@ -248,6 +256,42 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
...
+		return false
+	}
+	if !clusterExists {
+		responses.Send(
+			http.StatusNotFound,
+			writer,
+			(&storage.ItemNotFoundError{ItemID: clusterID}).Error(),
+		)
+		return false
+	}
```
Bee-lee: Why not return `storage.ItemNotFoundError` from storage and let `handleServerError` handle it? Or do you not like that approach?

server/server.go
```diff
@@ -248,6 +256,42 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
...
+
+	ruleExists, err := server.Storage.CheckIfRuleExists(ruleID)
+	if err != nil {
+		log.Err(err)
```
Bee-lee: Shouldn't it be `log.Error().Err(err)` ? Is it the same, just shorter?

server/server.go
```diff
@@ -248,6 +256,42 @@ func (server *HTTPServer) voteOnRule(writer http.ResponseWriter, request *http.R
...
+		responses.Send(http.StatusInternalServerError, writer, err.Error())
+		return false
+	}
+	if !ruleExists {
+		responses.Send(
```
Bee-lee: the returned err from responses.Send* is not handled

server/server.go
Sergey1011010: hm, maybe actually better, thanks

server/server.go
Sergey1011010: not sure

server/server.go
Sergey1011010: it's gonna be huge refactoring since we're not checking it everywhere :)

server/server.go
Sergey1011010: fixed

server/server.go
Bee-lee: we are, it was in my PR yesterday. That's also why the `handleServerError` function, so it doesn't have to be everywhere :D

server/server.go
Sergey1011010: where is this `handleServerError`?

Bee-lee: i guess you dont have master merged

Bee-lee: server/errors.go

Sergey1011010: nevermind, I'm using an old version of base branch

server/server.go
tisnik: +1

server/server.go
Sergey1011010: fixed

server/server.go
Sergey1011010: fixed

PR #407, opened 2020-03-27
[PoC] Monkey patching sarama functions

JoseLSegura:
# Description

Please, don't merge this PR yet. It is a prove of concept about using monkey patching for sarama library.
With monkey patches we can be able to substitute any `sarama` function (well, any Go function) and use a mocked version that returns whatever we want.

If you find this useful, I could write some utility methods that allow to set "expectations" instead of writing mock methods for every usage of every function, because it won't scale.

I have some concerns about using this "monkey" library due to some comments that I have read, for example:

> github.com/bouk/monkey makes assumptions about the structure of the compiled code that are not at all guaranteed by the language spec, so you should not expect it to continue to work in general.
[Source](https://github.com/golang/go/issues/27708#issuecomment-421997870).

If you agree and we can live with these cons, I will go ahead and start to add new utilities to being able to mock almost every function we want with expectations, if needed.

## Type of change

## Testing steps
I needed to add the `-gcflag` for running tests because monkey package doesn't work well with enabled inlining.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ run: clean build ## Build the project and executes the binary
 	./insights-results-aggregator
 
 test: clean build ## Run the unit tests
-	@go test -coverprofile coverage.out $(shell go list ./... | grep -v tests)
+	@go test -gcflags -l -coverprofile coverage.out $(shell go list ./... | grep -v tests)
 
 integration_tests: ## Run all integration tests
 	@echo "Running all integration tests"
--- a/consumer/consumer_test.go
+++ b/consumer/consumer_test.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/RedHatInsights/insights-results-aggregator/tests/testdata"
 
+	"bou.ke/monkey"
 	"github.com/Shopify/sarama"
 	mapset "github.com/deckarep/golang-set"
 	"github.com/stretchr/testify/assert"
@@ -377,6 +378,17 @@ func TestKafkaConsumer_New_MockBroker(t *testing.T) {
 		defer mockBroker.MustClose(t)
 		defer helpers.MustCloseConsumer(t, mockConsumer)
 	}, testCaseTimeLimit)
+}
+
+func TestMonkeyPatch(t *testing.T) {
+	guard := monkey.Patch(sarama.NewClient, helpers.NewMockClient)
+	defer guard.Unpatch()
+
+	consumer, err := consumer.NewWithSaramaConfig(broker.Configuration{}, nil, nil, false)
+
+	if consumer != nil || err == nil {
+		t.Fatal("WTF!")
+	}
 }
 
 func TestKafkaConsumer_New_MockBrokerNoTopicError(t *testing.T) {
--- a/tests/helpers/sarama_mocks.go
+++ b/tests/helpers/sarama_mocks.go
@@ -1,0 +1,16 @@
+package helpers
+
+import (
+	"errors"
+
+	"github.com/Shopify/sarama"
+)
+
+// func NewConsumerFromClient(client Client) (Consumer, error)
+func NewMockConsumerFromClient(client sarama.Client) (sarama.Consumer, error) {
+	return nil, errors.New("Monkey patched!!")
+}
+
+func NewMockClient(addrs []string, conf *sarama.Config) (sarama.Client, error) {
+	return nil, errors.New("Money patched!!")
+}


Bee-lee:
Can this even be considered testing though? I mean I understand the difficulty of actually testing this thing, and I don't have a better proposition, but I'm not sure if we can call it tested if we just mock the responses we want.

Pretty interesting package, though, and even more interesting code in that package..
```
// this function is super unsafe
// aww yeah
// It copies a slice to a raw memory location, disabling all memory protection before doing so.
func copyToLocation(location uintptr, data []byte) {
```

tisnik:
> Can this even be considered testing though? I mean I understand the difficulty of actually testing this thing, and I don't have a better proposition, but I'm not sure if we can call it tested if we just mock the responses we want.
> 
> Pretty interesting package, though, and even more interesting code in that package..
> 
> ```
> // this function is super unsafe
> // aww yeah
> // It copies a slice to a raw memory location, disabling all memory protection before doing so.
> func copyToLocation(location uintptr, data []byte) {
> ```

IMHO it must be ugly because it pushes Go technology to max. Better not to know what's under the roof :)

Sergey1011010:
Sounds great, but I agree with @Bee-lee that it's not really testing things. But it's of course better than nothing for the cases when we can't cover something with jocko/sarama mocks.

Sergey1011010:
Interesting package, it reminds me about buffer overflow in C(++) programs when you could rewrite return address on stack and call another function :)

Sergey1011010:
> Monkey won't work on some security-oriented operating system that don't allow memory pages to be both write and execute at the same time. With the current approach there's not really a reliable fix for this.

this could be a huge problem, we can't control the environment in CI, neither it's a good idea to disable such checks in your OS

JoseLSegura:
Well, it depends on what type on testing are you thinking about.

For sure, monkey patching is everything about unit testing. You are allowed to mock the boundaries of your code (in our case, `sarama` is not our code at all).

In Python world patches are used everywhere for testing. Some examples:

- You can patch opening and reading a file, because you don't want that your unit tests depends on the real file system.
- You can patch `socket` library in order to "send" and "receive" the byte streams you want for test some corner case and not depends on the real network at all
- You can patch a Kafka library because you want to get data "from Kafka" without deploying a Kafka server instance...

At the end, it always depends on the level of the testing you are willing to do. At the unit tests level, I think that patching/mocking external libraries is not bad. Obviously, in e2e testing is completely different, for sure.

Good debate!

JoseLSegura:
> > Monkey won't work on some security-oriented operating system that don't allow memory pages to be both write and execute at the same time. With the current approach there's not really a reliable fix for this.
> 
> this could be a huge problem, we can't control the environment in CI, neither it's a good idea to disable such checks in your OS

At the end, yes, we can control which image is used by Travis, for example. Using "go 1.13" is an election, and we can see what kind of OS is that container (I'm guessing that Travis is using docker for CI and, probably, a default golang container image. Probably I'm wrong)

Sergey1011010:
> > > Monkey won't work on some security-oriented operating system that don't allow memory pages to be both write and execute at the same time. With the current approach there's not really a reliable fix for this.
> > 
> > this could be a huge problem, we can't control the environment in CI, neither it's a good idea to disable such checks in your OS
> 
> At the end, yes, we can control which image is used by Travis, for example. Using "go 1.13" is an election, and we can see what kind of OS is that container (I'm guessing that Travis is using docker for CI and, probably, a default golang container image. Probably I'm wrong)

Even if we can choose an image, we still can't choose real operation system running this container

JoseLSegura:
> > > > Monkey won't work on some security-oriented operating system that don't allow memory pages to be both write and execute at the same time. With the current approach there's not really a reliable fix for this.
> > > 
> > > 
> > > this could be a huge problem, we can't control the environment in CI, neither it's a good idea to disable such checks in your OS
> > 
> > 
> > At the end, yes, we can control which image is used by Travis, for example. Using "go 1.13" is an election, and we can see what kind of OS is that container (I'm guessing that Travis is using docker for CI and, probably, a default golang container image. Probably I'm wrong)
> 
> Even if we can choose an image, we still can't choose real operation system running this container

Good point... I'm not sure if the memory pagination is performed by the container controller or by the real OS behind...

But at the end, I feel your fears :-D And I mostly agree with them, so forget about this weird library.

Thank you all for your comments!

Issue #383, opened 2020-03-24
[Fuzz testing] GET /organizations/2147483648/clusters returns 500


quarckster:
Response from the server:
```
{"status":"pq: value \"2147483648\" is out of range for type integer"}
````

tisnik:
This has been fixed already by Daniel, but we need to have proper tests for this issue as well. So I'm continuing working on this task.

PR #414, opened 2020-03-30
Reproducer for issue #383

tisnik:
# Description

Reproducer for issue #383

## Type of change

- REST API tests

## Testing steps

You can run `make rest_api_tests` on command line


tisnik:
> Aren't we running this make rule into CI?

we do. why?

Issue #417, opened 2020-03-30
Make rules parser accept new changes


Bee-lee:
- There is a new field in the structure of rules content: tags - list of string keys that will be used for rule grouping

- There is also a new change implemented in the content, regarding the `impact` field, where instead of a numeric value, it is a string key that has to be matched according to the defined value

Both the list of tags and the impact keyword values are located in a config file for rules-content here https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp/blob/master/content/config.yaml. There is a related issue #418 regarding the loading of this file from the repository alongside the content.

PR #420, opened 2020-03-30
add a way to update local rules content for testing purposes

### LINKED ISSUE #419: Add a way to update local rules-content for testing purposes
As a developer I want a simple way to update my rules-content locally since the Dockerfile that normally does this is not intended for local use

Bee-lee:
# Description
I added a very simple (I'd like to say dumb) script that updates the `./tests/content/ok` directory with rules currently present in the [rules content directory](https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp/tree/master/content/external/rules).

This is just for local use as of now, you need to be in VPN to access that repo... so it's far from perfect, but you can use this to quickly test against real rules and with modifying the `INSIGHTS_RESULTS_AGGREGATOR__CONTENT__PATH` to use normally.

Right now, I'm not including the new content in the OK directory, because the rules parser currently needs some changes -- see #417 , but it can be used by @natiiix or whoever will work on that to test this.

The script also copies the content configuration file `config.yaml` needed to work on that task.

pls let me know if you have a better solution, I didnt want to spend too much time on it

Fixes #419

## Type of change

## Testing steps
`make before_commit` without updating the rules.

### DIFF
--- a/update_rules_content.sh
+++ b/update_rules_content.sh
@@ -21,10 +21,10 @@ function clean_up() {
 trap clean_up EXIT
 
 RULES_REPO="https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp.git"
-CONTENT_DIR="$( cd "$(dirname "$0")" && pwd )/tests/content"
-OK_RULES_DIR="${CONTENT_DIR}/ok"
+SCRIPT_DIR="$(dirname $(realpath $0))"
+CONTENT_DIR="${SCRIPT_DIR}/rules-content"
 
-CLONE_TEMP_DIR="${CONTENT_DIR}/tmp"
+CLONE_TEMP_DIR="${SCRIPT_DIR}/.tmp"
 EXTERNAL_RULES_CONTENT="${CLONE_TEMP_DIR}/content/external/rules/"
 RULES_CONFIG="${CLONE_TEMP_DIR}/content/config.yaml"
 
@@ -37,7 +37,7 @@ then
     exit 1
 fi
 
-rm -rf "$OK_RULES_DIR"
+rm -rf "${CONTENT_DIR}"
 
 if [ $? -ne 0 ]
 then
@@ -45,7 +45,7 @@ then
     exit 1
 fi
 
-mv "${EXTERNAL_RULES_CONTENT}" "${OK_RULES_DIR}" && mv "${RULES_CONFIG}" "${OK_RULES_DIR}"
+mv "${EXTERNAL_RULES_CONTENT}" "${CONTENT_DIR}" && mv "${RULES_CONFIG}" "${CONTENT_DIR}"
 
 if [ $? -ne 0 ]
 then
@@ -55,5 +55,5 @@ fi
 
 rm -rf "${CLONE_TEMP_DIR}"
 
-echo "tests/content/ok/ updated with latest rules"
+echo "${CONTENT_DIR} updated with latest rules"
 exit 0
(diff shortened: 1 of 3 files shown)


update_rules_content.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# Updates the ./tests/content/ok/ directory with the latest rules to test against
+
+function clean_up() {
+    rm -rf "$CLONE_TEMP_DIR"
+}
+trap clean_up EXIT
+
+RULES_REPO="https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp.git"
+CONTENT_DIR="$( cd "$(dirname "$0")" && pwd )/tests/content"
```
Sergey1011010: isn't `./tests/content` enough here?

update_rules_content.sh
```diff
@@ -0,0 +1,59 @@
...
+OK_RULES_DIR="${CONTENT_DIR}/ok"
```
Sergey1011010: Now some tests use `./tests/content/ok` so maybe it's better to use another directory so that it wouldn't break any tests if content changes

update_rules_content.sh
JoseLSegura: Or just `$(dirname "$0")`, no need to "cd" and "pwd", I guess

Bee-lee: @Sergey1011010 @JoseLSegura Can you tell me what are the advantages of using relative paths? I don't like them, so I use absolute paths whenever I can. 

`./tests/content` doesn't work when you call the script from another location, e.g. from the ok/ folder when I was testing it
`$(dirname "$0")` works when called from different location, but the outputted messages are less readable for me, e.g. `Attempting to clone repository into ../../../tests/content/tmp` instead of the readable absolute path

update_rules_content.sh
Bee-lee: @Sergey1011010 Isn't that what we want?? For our tests to pass against real data?

update_rules_content.sh
JoseLSegura: Well, if you prefer absolute path I'm ok with that, but it's better to use `realpath` instead of "cd + pwd", IMHO.

In any case, you can merge from my point of view

natiiix: The readability of the code kinda suffers from using complicated/nested commands like this.

However, the much more pressing issue is why are we even putting real rule content into the `tests` directory? As far as understand, the rules are secret, but the tests directory is completely publicly visible. Or am I reading the script wrong and is that not what's happening here?

Bee-lee: I don't understand how is `"$( cd "$(dirname "$0")" && pwd )"` complicated or unreadable??
I can change it, but then the readability of the output is impacted. 

Regarding your second issue, I had only one idea how to solve and that would be to put it in a folder that would be in `.gitignore`, but I thought that testing against real rules would be a good idea instead of tailoring our mocked data. I didn't realize the content of the rules is to be kept secret, @tisnik can you weigh in on this issue pls?

natiiix: @Bee-lee Using `cd` and two levels of nested commands, which requires an unforunate number of quotation marks, certainly isn't the most readable code. However, that's a general problem with any shell script code. The alternative with `realpath` seems easier to comprehend to me.

Well, the rule content is not available anywhere publicly and there is no plan for it to be, so I'd say we definitely want to keep certain information about the rules and their content private, from what I've heard so far.

Bee-lee: @natiiix take a look now, i change it to realpath, changed the content directory and added it to .gitignore

Bee-lee:
Thank you for your comments guys!

Issue #418, opened 2020-03-30
Include ccx-rules-ocp/content/config.yaml in the Docker image and the local rules updater to be parsed


Bee-lee:
Currently there are some new changes implemented in the rules content https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp/tree/master/content/external, which need additional information from the content configuration located at https://gitlab.cee.redhat.com/ccx/ccx-rules-ocp/blob/master/content/config.yaml.

This configuration file has to included in the Docker image and the local rules updater so that it can be parsed by the content parser.

- [ ] include the `config.yaml` in the Docker image
- [ ] include the `config.yaml` in the local rules updating script

Bee-lee:
Closing this issue, because the content parser is now expecting the whole content/ directory, instead of separate rules + config file

PR #436, opened 2020-03-31
Update content parser to match new requirements

### LINKED ISSUE #417: Make rules parser accept new changes

### LINKED ISSUE #426: Convert impact values stored in OCP rules from strings to integers
## Problem description

The following PR is to be merged into OCP rules repository:
The impact shouldn't be a number, but a value from `ccx-rules-ocp/content/config.yaml`

It means that we'd need to convert the string back to integer. We might use mapping stored in `config.yaml`.

## Additional information:

Configuration with mapping:
OCP Git repository - ccx-rules-ocp/content/config.yaml

Pull requests with changes in OCP rules repository:
#69

natiiix:
# Description

The global `config.yaml` containing the `impact` map is now parsed and used before inserting  impact values into database.

Tags are now parsed for each rule content error key. We're not gonna use this for now yet, but I already had it done, so I included it in this PR.

I have tried to also update the `Dockerfile`, but I am not 100% sure I did that right.

Fixes #426 
Fixes #417 
Partially contributes towards #418 

## Type of change

## Testing steps

`make before_commit`


content/content_test.go
```diff
@@ -33,34 +33,43 @@ func TestContentParseOK(t *testing.T) {
...
 	assert.True(t, exists, "'err_key' error content is missing")
 }
 
+// TestContentParseOKNoContent checks that parsing content when there is no rule
+// content available, but the file structure is otherwise okay, succeeds.
+func TestContentParseOKNoContent(t *testing.T) {
+	con, err := content.ParseRuleContentDir("../tests/content/ok_no_content/")
+	helpers.FailOnError(t, err)
+
+	assert.Equal(t, 0, len(con.Rules))
```
Sergey1011010: It would be better to use assert.Empty here

natiiix: @Sergey1011010 Good point, didn't know it existed.

PR #437, opened 2020-03-31
update Dockerfile to include rules content config file

Bee-lee:
# Description
Include config.yaml for content rules to be parsed by rules parser.

## Type of change

## Testing steps
built the docker image with the gitlab repo, since this Dockerfile isnt intended for local use, run it and checked that the files + config file is present


Bee-lee:
@natiiix as previously discussed :D

PR #439, opened 2020-03-31
Match parser with rules content update script + update README

Bee-lee:
# Description
- updated `update_rules_content.sh` to conform with latest content parser changes
- updated README.md with information about rules content

## Type of change

- Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,42 @@ CREATE TABLE report (
 )
 ```
 
+### Tables rule and rule_error_key
+
+These tables represent the content for Insights rules to be displayed by OCM.
+The table `rule` represents more general information about the rule, whereas the `rule_error_key`
+contains information about the specific type of error which occured. The combination of these two create a unique rule.
+Very trivialized example could be:
+- rule "REQUIREMENTS_CHECK"
+    - error_key "REQUIREMENTS_CHECK_LOW_MEMORY"
+    - error_key "REQUIREMENTS_CHECK_MISSING_SYSTEM_PACKAGE"
+
+```sql
+CREATE TABLE rule (
+    module        VARCHAR PRIMARY KEY,
+    name          VARCHAR NOT NULL,
+    summary       VARCHAR NOT NULL,
+    reason        VARCHAR NOT NULL,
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,14 +99,20 @@ CREATE TABLE report (
 -- -1 is dislike
 CREATE TABLE cluster_rule_user_feedback (
     cluster_id VARCHAR NOT NULL,
-    rule_id INTEGER  NOT NULL,
+    rule_id VARCHAR NOT NULL,
```
tisnik: that's interesting how it worked before?

README.md
```diff
@@ -179,6 +221,27 @@ or you can use integration tests suite. More details are [here](https://gitlab.c
...
+
+The __hit__ rules are the rules that the customer is interested in and therefore the information about them (their __content__) needs to be displayed in OCM. The content for the rules is present in another repository, alongside the actual rule implementations, which is primarily maintained by the support-facing team.
+For that reason, the `insights-results-aggregator` is setup in a way, that the content we're interested in is copied to the Docker image from the repository during image build and we have a push webhook on master branch set up in that repository, signalling our app to be rebuilt.
+
+This content is then processed upon application start-up, correctly parsed by the package `content` and saved into the database (see DB structure).
+
+When a request for a cluster report comes from OCM, the report is parsed (TODO: parse reports only once when consuming them) and content for all the hit rules is returned.
+
```
tisnik: very nice explanation!!!

README.md
Sergey1011010: Because we use sqlite and don't test on postgres ;)

Bee-lee: This is from the new recent migration where the schema was altered, just the readme wasnt updated, so it was as OK then

Sergey1011010: oh, actually no, the migration is corrent, only readme was outdated

tisnik: > Because we use sqlite and don't test on postgres ;)

yeah that's Visual Basic approach with its stupid `VARIANT` type

README.md
Bee-lee: TYVM!

Issue #447, opened 2020-04-02
Ability to anonymize input data in order to have a good database for testing


tisnik:
We need to be able to anonymize input data in order to have a good database for testing. Simple Python script would be sufficient.

PR #448, opened 2020-04-02
Script to anonymize input

tisnik:
# Description

Script to anonymize input data. Needed for being able to test aggregator and its UI correctly.

Fixes issue #447 

## Type of change

- Documentation update

## Testing steps

N/A

### DIFF
--- a/utils/anonymize.py
+++ b/utils/anonymize.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+
+# Copyright © 2020 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


tisnik:
> Where is it supposed to be executed? Inside the container? Then you need to update the Dockerfile.

on devel machines. It is just for data preparation.

utils/anonymize.py
```diff
@@ -0,0 +1,49 @@
...
+
+            # replace content of "info" node
+            if "info" in data:
+                data["info"] = []
+
+            # generate output
+            outfilename = "s_{:0>5d}.json".format(i)
+            i += 1
+            with open(outfilename, "w") as fout:
+                json.dump(data, fout, indent=4)
```
JoseLSegura, line 49: All of this should be outside the `with` context, because you don't need to keep the file opened, but this is a very silly problem, so you can ignore this comment if you want

PR #471, opened 2020-04-03
Print version info on startup

### LINKED ISSUE #470: Print version info, commit hash etc. on startup
Print version info, commit hash etc. on startup to allow us to identify problems related with older versions

tisnik:
# Description

Print version info on startup

Fixes #470 

## Type of change

## Testing steps

Just `make build` and start aggregator.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# Copyright 2020 Red Hat, Inc
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# this is improper - we need to start using tags in GitHub properly
+version=0.5
+
+buildtime=$(date)
+branch=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# this is improper - we need to start using tags in GitHub properly
+version=0.5
+
+buildtime=$(date)
+branch=$(git rev-parse --abbrev-ref HEAD)
+commit=$(git rev-parse HEAD)
+
+go build -ldflags="-X 'main.BuildTime=$buildtime' -X 'main.BuildVersion=$version' -X 'main.BuildBranch=$branch' -X 'main.BuildCommit=$commit'"
+exit $?
```
quarckster, line 24: How will this work with the current deployment process?

tisnik:
> LGTM
> 
> I think we need to do the same for ccx-data-pipeline. Not as important as it doesn't have an interface to other services, but we should do this as well

actually it would be much easier in Python, no need to play with `ldflags` etc.

build.sh
JoseLSegura: Mmmm good point. The Travis and Codecov apparently have git in their environments, but IDK if it will be the case on Jenkins to handle the stable branch...

quarckster: @JoseLSegura it should work fine. Jenkins doesn't build the code for that repo. It operates images.

JoseLSegura: I was concerned about the container image generation process, but I have reviewed the Dockerfile and I saw that we already need `git` to generate the container, so no problems at all

JoseLSegura:
> > LGTM
> > I think we need to do the same for ccx-data-pipeline. Not as important as it doesn't have an interface to other services, but we should do this as well
> 
> actually it would be much easier in Python, no need to play with `ldflags` etc.

I opened an enhancement issue on ccx-data-pipeline to discuss there.

TY!

PR #474, opened 2020-04-03
Script to fill in results

### LINKED ISSUE #473: Complete script for filling aggregator db with anonymized real data
Currently we play with mocked data in our database, we need to have a real data set that we can show to our customers. The goal of this task is to download the data from production and fill the aggregator database in the production pipeline with it so that we can see what the customers see.

tisnik:
# Description

Script to fill in results

Fixes #473

## Type of change

## Testing steps

I can show you :)


utils/fill_in_results.sh
```diff
@@ -0,0 +1,147 @@
...
+then
+    echo -e "${BLUE}Usage:${NC}"
+    echo "    fill_in_results.sh archive.tar.bz org_id cluster_name"
+    echo -e "${BLUE}Example:${NC}"
+    echo "    fill_in_results.sh external-rules-archives-2020-03-31.tar 11789772 5d5892d3-1f74-4ccf-91af-548dfc9767aa"
+    exit 1
+fi
+
+check_insights_availability
+decompress_input exernal-rules-archives-2020-03-31.tar
```
Bee-lee, line 139: LGTM so far, I assume this will be changed to the param?

utils/fill_in_results.sh
```diff
@@ -0,0 +1,147 @@
...
+# We just move all files from this structure into one "flat" workdir
+function move_input_to_workdir() {
+    find . -name "*.tar.gz" -exec mv -i -- {} "$1" \;
+}
+
+
+# Remove now empty temporary directories
+function cleanup_temp_directories() {
+    # shellcheck disable=SC2045,SC2035
+    for temp_directory in $(ls -1 -d */)
```
Sergey1011010, line 51: hope nobody will add a space between `*` and `/` :)

utils/fill_in_results.sh
```diff
@@ -0,0 +1,147 @@
...
+        else
+            echo -e "${BLUE}Skipping ${temp_directory}${NC}"
+        fi
+    done
+}
+
+# Apply all OCP rules for all input files (gzipped tarballs)
+# Output is formatted to be human readable
+function apply_ocp_rules() {
+    cwd=$(pwd)
```
Sergey1011010, line 66: isn't is the same as $PWD?

utils/fill_in_results.sh
tisnik: definitely, yes, $1

utils/fill_in_results.sh
tisnik: yup, and run it with sudo :)

PR #475, opened 2020-04-03
Record consumer errors in storage

### LINKED ISSUE #353: Store improper consumed messages into a form of DLQ
As a devops I'd like to have a better information about consumed messages that can't be processed for any reason. Those messages (it's content + metadata) might be stored in some form of DLQ, probably directly in storage (separate table) to be analyzed later.

natiiix:
# Description

When the consumer is unable to process a message from Kafka, it will log the error and then it will add a record to the storage. This record includes all information about the consumed message, the error message and when the incident happened.

Fixes #353 

## Type of change

## Testing steps

`make before_commit` (still without the OpenAPI part)

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/DATA-DOG/go-sqlmock"
 	"github.com/RedHatInsights/insights-results-aggregator/tests/testdata"
+	"github.com/Shopify/sarama"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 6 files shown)


storage/storage_test.go
```diff
@@ -578,3 +579,48 @@ func TestDBStorage_NewSQLite(t *testing.T) {
...
+	err := mockStorage.WriteConsumerError(&sarama.ConsumerMessage{
+		Topic:     testTopic,
+		Partition: testPartition,
+		Offset:    testOffset,
+		Key:       testKey,
+		Value:     testMessage,
+		Timestamp: testProducedAt,
+	}, testError)
+
+	assert.NoError(t, err)
```
Sergey1011010, line 604: it's better to use `helpers.FailOnError(t, err)`

natiiix: @Sergey1011010 Good point. I didn't realize `NoError` didn't immediately fail the test.

PR #478, opened 2020-04-03
Endpoints for integration tests

### LINKED ISSUE #476: Add debug only endpoint to fill rule related tables
For some integrations tests(for example like/dislike endpoints) it's necessary to have a rule in the database.

### LINKED ISSUE #477: Add debug only endpoint to get user's vote on rule
It's necessary for integration tests to be sure like/dislike endpoints work.

Sergey1011010:
# Description

Fixes #476
Fixes #477

## Type of change

## Testing steps

`make before_commit`


server/server.go
```diff
@@ -247,55 +247,153 @@ func (server *HTTPServer) checkVotePermissions(writer http.ResponseWriter, reque
...
+
+	err = responses.SendResponse(writer, responses.BuildOkResponseWithData(
+		"rule_error_key", ruleErrorKey,
+	))
 	if err != nil {
 		log.Error().Err(err).Msg(responseDataError)
 	}
 }
 
+func (server *HTTPServer) readVoteOnRuleParams(writer http.ResponseWriter, request *http.Request) (types.ClusterName, types.RuleID, types.UserID, bool) {
```
bond95, line 356: I think we should extract functions like this to separate file, and in `server.go` leave only endpoints

tisnik: orgId don't have to be handled. It's per-user vote

bond95: @tisnik in that case how do we check permissions? What if user orgId and cluster orgId don't match?

server/server.go
```diff
@@ -247,55 +247,153 @@ func (server *HTTPServer) checkVotePermissions(writer http.ResponseWriter, reque
...
+		return "", "", "", false
+	}
+
+	_, err = server.Storage.GetRuleByID(ruleID)
+	if err != nil {
+		handleServerError(writer, err)
+		return "", "", "", false
+	}
+
+	err = server.checkVotePermissions(writer, request, clusterID)
```
bond95, line 388: I found it.

server/server.go
bond95: Found it

Sergey1011010: done

PR #485, opened 2020-04-06
fixed offset column in conumser error table

### LINKED ISSUE #484: Rename `offset` column in consumer_error table
It causes a lot of issues in some queries because postgres can't get that it's not a keyword, but a column name

Sergey1011010:
# Description

Fixes #484

## Type of change

## Testing steps

Run with postgres and see that consumer errors are written into the table and `make before_commit`

### DIFF
--- a/migration/migration5.go
+++ b/migration/migration5.go
@@ -24,16 +24,18 @@ var mig5 = Migration{
 	StepUp: func(tx *sql.Tx) error {
 		_, err := tx.Exec(`
 			CREATE TABLE consumer_error (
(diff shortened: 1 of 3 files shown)


Sergey1011010:
@natiiix good catch, thanks, I'll reopen the issue and will fix it properly https://github.com/RedHatInsights/insights-results-aggregator/issues/484

Sergey1011010:
@natiiix, actually, previous schema was failing on postgres, so we can't have anything with old migrations deployed.

PR #515, opened 2020-04-08
fixed no body errors in debug endpoints

### LINKED ISSUE #513: create rule endpoint returns 500 error when body is empty
```bash
curl -v -X POST http://localhost:8080/api/insights-results-aggregator/v1/rules/some.python.module
```
Current behaviour:
```
< HTTP/1.1 500 Internal Server Error
{"status":"Internal Server Error"}
```
Expected behaviour: 400 error with some message

### LINKED ISSUE #514: create rule_error_key endpoint returns 500 error when body is empty
```bash
curl -v -X POST http://localhost:8080/api/insights-results-aggregator/v1/rules/some.python.module/error_keys/ek
```
Current behaviour:
```
< HTTP/1.1 500 Internal Server Error
{"status":"Internal Server Error"}
```
Expected behaviour: 400 error with some message

Sergey1011010:
# Description

Fixes #513
Fixes #514

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -75,7 +75,7 @@ func handleServerError(writer http.ResponseWriter, err error) {
 
 	switch err := err.(type) {
 	case *RouterMissingParamError, *RouterParsingError, *json.SyntaxError, *NoBodyError:
-		respErr = responses.Send(http.StatusBadRequest, writer, err.Error())
+		respErr = responses.SendError(writer, err.Error())
 	case *storage.ItemNotFoundError:
 		respErr = responses.SendNotFound(writer, err.Error())
 	case *AuthenticationError:
--- a/server/errors.go
+++ b/server/errors.go
@@ -60,6 +60,13 @@ func (e *AuthenticationError) Error() string {
 	return e.errString
 }
 
+// NoBodyError error meaning that client didn't provide body when it's required
+type NoBodyError struct{}
+
+func (*NoBodyError) Error() string {
+	return "client didn't provide request body"
+}
+
 // handleServerError handles separate server errors and sends appropriate responses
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -295,6 +295,9 @@ func (server *HTTPServer) createRule(writer http.ResponseWriter, request *http.R
 
 	err = json.NewDecoder(request.Body).Decode(&rule)
 	if err != nil {
+		if err.Error() == "EOF" {
```
tisnik, line 298: is not it better to use constant?

server/errors.go
```diff
@@ -60,23 +60,26 @@ func (e *AuthenticationError) Error() string {
...
 
 	var respErr error
 
 	switch err := err.(type) {
-	case *RouterMissingParamError:
-		respErr = responses.SendError(writer, err.Error())
-	case *RouterParsingError:
-		respErr = responses.SendError(writer, err.Error())
+	case *RouterMissingParamError, *RouterParsingError, *json.SyntaxError, *NoBodyError:
+		respErr = responses.Send(http.StatusBadRequest, writer, err.Error())
```
Bee-lee, line 78: Can you pls use the `SendError()` so that its unified? `Send()` is currently only used here and I will rename it in `utils` next sprint to `SendBadRequest()`

server/server.go
Sergey1011010: good idea, thanks

server/errors.go
Sergey1011010: The reason why I didn't use SendError() is that it's not really clear from the name of the function what kind of error it's gonna send. 

> I will rename it in utils next sprint to SendBadRequest()

Okay, thanks

PR #520, opened 2020-04-08
fixed create_rule and create_rule_error_key endpoints 404 and 500 errors

### LINKED ISSUE #516: create rule endpoint returns 500 error when body has json with wrong structure
```bash
curl -v -X POST http://localhost:8080/api/insights-results-aggregator/v1/rules/some.python.module --data '{"module": []}'
```
Current behaviour:
```
< HTTP/1.1 500 Internal Server Error
{"status":"Internal Server Error"}
```
Expected behaviour: 400 error with description what's happened

### LINKED ISSUE #517: create rule_error_key endpoint returns 500 error when body has json with wrong structure
```bash
curl -v -X POST http://localhost:8080/api/insights-results-aggregator/v1/rules/some.python.module/error_keys/ek --data '{"rule_module": []}'
```
Current behaviour:
```
< HTTP/1.1 500 Internal Server Error
{"status":"Internal Server Error"}
```
Expected behaviour: 400 error with description what's happened

### LINKED ISSUE #519: create rule_error_key endpoint should return 404 when rule does not exist

Sergey1011010:
# Description

Fixes #516
Fixes #517
Fixes #519

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -76,6 +76,8 @@ func handleServerError(writer http.ResponseWriter, err error) {
 	switch err := err.(type) {
 	case *RouterMissingParamError, *RouterParsingError, *json.SyntaxError, *NoBodyError:
 		respErr = responses.SendError(writer, err.Error())
+	case *json.UnmarshalTypeError:
+		respErr = responses.SendError(writer, "bad type in json data")
 	case *storage.ItemNotFoundError:
 		respErr = responses.SendNotFound(writer, err.Error())
 	case *AuthenticationError:
--- a/server/server.go
+++ b/server/server.go
@@ -345,6 +345,13 @@ func (server *HTTPServer) createRuleErrorKey(writer http.ResponseWriter, request
 		return
 	}
 
+	// it's gonna raise an error if rule does not exist
+	_, err = server.Storage.GetRuleByID(ruleID)
+	if err != nil {
+		handleServerError(writer, err)
+		return
+	}
+
 	errorKey, err := readErrorKey(writer, request)
 	if err != nil {
 		// everything has been handled already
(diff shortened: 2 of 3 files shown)


server/errors.go
```diff
@@ -76,6 +76,8 @@ func handleServerError(writer http.ResponseWriter, err error) {
 	switch err := err.(type) {
 	case *RouterMissingParamError, *RouterParsingError, *json.SyntaxError, *NoBodyError:
 		respErr = responses.SendError(writer, err.Error())
+	case *json.UnmarshalTypeError:
+		respErr = responses.SendError(writer, "bad type in json data")
```
Bee-lee, line 80: LGTM overall, is bad type the only error this can mean?

Sergey1011010: I guess, because it's called UnmarhsalTypeError :)

Bee-lee: @Sergey1011010 I mean, why not the err.Error()?

Sergey1011010: @Bee-lee err.Error() would reveal some data about internal types. The project is open-sourced, but still. @tisnik what do you think?

Bee-lee: @Sergey1011010 yeah I'm not sure in this case, when I was doing the refactoring, I replaced the messages and err.Errors for the InternalServerErrors, which can definitely reveal some stuff, so I see what you mean, but I don't know, this is still a client error

tisnik: IMHO it is ok, logs are stored quite securely :-)

tisnik: it is about types, not actual data, right?

Sergey1011010: @tisnik I mean that if we return err.Error() to the user it would show message which would contain info about about type, something like:
```
cannot unmarshal array into Go struct field TypeName.FieldName of type string
```

PR #530, opened 2020-04-09
update Dockerfile and rule updating script to expect tutorial rule content

Bee-lee:
# Description
ssia

## Type of change

## Testing steps
tested Dockerfile locally with credentials for Gitlab + checked image
make before_commit


Dockerfile
```diff
@@ -37,6 +37,8 @@ COPY --from=builder /go/insights-results-aggregator/insights-results-aggregator
 COPY --from=builder /go/insights-results-aggregator/openapi.json /openapi/openapi.json
 # copy just the rule content instead of the whole repository
 COPY --from=builder /rules-content/content/ /rules-content
+# copy tutorial/fake rule hit on all reports
+ADD rules/tutorial/content/ /rules-content/external/rules
```
quarckster, line 41: Why do you use ADD instead of COPY?

Bee-lee: good question... :D changed it

PR #562, opened 2020-04-16
Search for rule content recursively

natiiix:
# Description

Apparently we should be able to search for rule content recursively while parsing it and this change aims to achieve precisely that.

## Type of change

## Testing steps

`make before_commit` without the OpenAPI test, which still keeps on crashing.

### DIFF
--- a/content/content.go
+++ b/content/content.go
@@ -19,6 +19,7 @@ package content
 
 import (
 	"io/ioutil"
+	"os"
 	"path"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


natiiix:
> It looks ok, thank you @natiiix, but ... do we really need it?

I hope this will prevent the otherwise inevitable "new rule content has been added, but it's not presented to the user" issue caused by an unexpected, nested rule content somewhere deeper than we used to look. It has been suggested two weeks ago, I believe, that this can and likely will happen in the future.

PR #569, opened 2020-04-17
Random payload generator

### LINKED ISSUE #568: Enhancement of the script to generate semi-random payload
Expected new features:
- add items with random content
- replace existing content with random one

Pls.note: random content is not just 'strings', but other JS-relevant data types.

tisnik:
# Description

Types of input message mutation:
* any item (identified by its key) can be removed
* new items with random key and content can be added - new functionality
* any item can be replaced by new random content - new functionality

Fixes #568 

## Type of change

### DIFF
--- a/utils/random_payload_generator.py
+++ b/utils/random_payload_generator.py
@@ -21,12 +21,14 @@ import random
 class RandomPayloadGenerator:
     """Generator of random payload for testing API."""
 
-    def __init__(self):
+    def __init__(self, max_iteration_deep=2, max_dict_key_length=5, max_list_length=5,
+                 max_string_length=10):
         """Initialize the random payload generator."""
         self.iteration_deep = 0
-        self.max_iteration_deep = 2
-        self.max_dict_key_length = 10
-        self.max_string_length = 20
+        self.max_iteration_deep = max_iteration_deep
+        self.max_list_length = max_list_length
+        self.max_dict_key_length = max_dict_key_length
+        self.max_string_length = max_string_length
         self.dict_key_characters = string.ascii_lowercase + string.ascii_uppercase + "_"
         self.string_value_characters = (string.ascii_lowercase + string.ascii_uppercase +
                                         "_" + string.punctuation + " ")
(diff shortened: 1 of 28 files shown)


utils/random_payload_generator.py
```diff
@@ -0,0 +1,108 @@
...
+    def generate_random_list_or_string(self):
+        """Generate list filled in with random strings."""
+        if self.iteration_deep < self.max_iteration_deep:
+            self.iteration_deep += 1
+            value = self.generate_random_list(5)
+            self.iteration_deep -= 1
+        else:
+            value = self.generate_random_value(str)
+        return value
+
```
JoseLSegura, line 74: I don't understand this piece of code. `self.max_iteration_deep` is always 2, and `self.iteration_deep` starts as 0...
Taking into account that, when you enter in the `if` branch you will increment and then decrement the value of `self.iteration_deep`, I don't see that it is possible to enter in the `else` branch at all.

So, `generate_random_list_or_string` always will generate a random list with 5 as length.

utils/random_payload_generator.py
```diff
@@ -0,0 +1,108 @@
...
+    def generate_random_dict_or_string(self):
+        """Generate dict filled in with random strings."""
+        if self.iteration_deep < self.max_iteration_deep:
+            self.iteration_deep += 1
+            value = self.generate_random_dict(5)
+            self.iteration_deep -= 1
+        else:
+            value = self.generate_random_value(str)
+        return value
```
JoseLSegura, line 83: Same as before, the `else` branch is never reached

utils/random_payload_generator.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self):
+        """Initialize the random payload generator."""
+        self.iteration_deep = 0
+        self.max_iteration_deep = 2
+        self.max_dict_key_length = 10
+        self.max_string_length = 20
+        self.dict_key_characters = string.ascii_lowercase + string.ascii_uppercase + "_"
+        self.string_value_characters = (string.ascii_lowercase + string.ascii_uppercase +
+                                        "_" + string.punctuation + " ")
```
JoseLSegura, line 32: This variables are never used (at least internally in this class)

utils/random_payload_generator.py
tisnik: the script is more universal and all attrbiutes can be changed in future, if needed

PR #586, opened 2020-04-17
Cloudwatch config

### LINKED ISSUE #518: Add configuration for logging to CloudWatch
To being able to send logs to CloudWatch, we need to get some configuration about AWS authentication, log group and stream name.
We need to configure it from environment variables and/or configuration file

JoseLSegura:
# Description

Adding configuration and getters for CloudWatch configuration. It allows to configure both from configuration file and environment variables.

Updated config.toml and config-devel.toml too to put an example of usage of the new configuration.

Fixes #518 

## Type of change

## Testing steps
CI


logger/configuration.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Configuration for CloudWatch logger
+package logger
+
+// Configuration represents configuration of CloudWatch logger
+type Configuration struct {
+	AWS_access_id  string `mapstructure:"aws_access_id" toml:"aws_access_id"`
```
Sergey1011010, line 22: I think it's better to use camel case here :)

config.toml
```diff
@@ -31,3 +31,10 @@ pg_port = 5472
 pg_db_name = "aggregator"
 pg_params = ""
 log_sql_queries = true
+
+[cloudwatch]
+aws_access_id = "a key id"
+aws_secret_key = "tshhhh it is a secret"
+aws_region = "us-east-1"
+log_group = "platform-dev"
+stream_name = "insights-results-aggregator"
```
natiiix, line 40: Just a tiny note: missing final newline

JoseLSegura: On it

Issue #577, opened 2020-04-17
Organization whitelist path from config.toml is ignored


Bee-lee:
Because the `config` struct in configuration.go doesn't reflect the config structure.

The `mapstructure`/`toml` key is defined as `org_white_list_file`, which means that this check https://github.com/RedHatInsights/insights-results-aggregator/blob/master/configuration.go#L117 will always pass and the actual path that is used every time is actually `defaultOrgWhiteListFileName = "org_whitelist.csv"` 

Will be fixed alongside #489

PR #587, opened 2020-04-17
Optional organization whitelisting

### LINKED ISSUE #489: Ability to enable or disable organization whitelisting via configuration file
We need to be able to enable or disable organization whitelisting via configuration file.

Bee-lee:
# Description
Added option to enable whitelisting, since we dont want it now. I'm not deleting it yet because the requirements might change again really quick :) 
Also fixed a bug where the path defined by `org_whitelist` was always ignored. (also fixed in e2e)

Fixes #489 #577 

## Type of change

- Documentation update

## Testing steps
`make before_commit`, locally tried `INSIGHTS_RESULTS_AGGREGATOR__PROCESSING__ORG_WHITELIST=` variable

### DIFF
--- a/broker/configuration.go
+++ b/broker/configuration.go
@@ -29,7 +29,7 @@ type Configuration struct {
 	PublishTopic        string     `mapstructure:"publish_topic" toml:"publish_topic"`
 	Group               string     `mapstructure:"group" toml:"group"`
 	Enabled             bool       `mapstructure:"enabled" toml:"enabled"`
-	OrgWhitelist        mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
+	OrgWhitelist        mapset.Set `mapstructure:"org_whitelist_file" toml:"org_whitelist_file"`
 	OrgWhitelistEnabled bool       `mapstructure:"enable_org_whitelist" toml:"enable_org_whitelist"`
 	SaveOffset          bool       `mapstructure:"save_offset" toml:"save_offset"`
 }
--- a/broker/configuration.go
+++ b/broker/configuration.go
@@ -24,11 +24,12 @@ import (
 
 // Configuration represents configuration of Kafka broker
 type Configuration struct {
-	Address      string     `mapstructure:"address" toml:"address"`
-	Topic        string     `mapstructure:"topic" toml:"topic"`
-	PublishTopic string     `mapstructure:"publish_topic" toml:"publish_topic"`
-	Group        string     `mapstructure:"group" toml:"group"`
-	Enabled      bool       `mapstructure:"enabled" toml:"enabled"`
-	OrgWhitelist mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
-	SaveOffset   bool       `mapstructure:"save_offset" toml:"save_offset"`
+	Address             string     `mapstructure:"address" toml:"address"`
+	Topic               string     `mapstructure:"topic" toml:"topic"`
+	PublishTopic        string     `mapstructure:"publish_topic" toml:"publish_topic"`
+	Group               string     `mapstructure:"group" toml:"group"`
+	Enabled             bool       `mapstructure:"enabled" toml:"enabled"`
+	OrgWhitelist        mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
+	OrgWhitelistEnabled bool       `mapstructure:"enable_org_whitelist" toml:"enable_org_whitelist"`
+	SaveOffset          bool       `mapstructure:"save_offset" toml:"save_offset"`
 }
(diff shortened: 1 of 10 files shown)


broker/configuration.go
```diff
@@ -24,11 +24,12 @@ import (
...
-	Enabled      bool       `mapstructure:"enabled" toml:"enabled"`
-	OrgWhitelist mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
-	SaveOffset   bool       `mapstructure:"save_offset" toml:"save_offset"`
+	Address             string     `mapstructure:"address" toml:"address"`
+	Topic               string     `mapstructure:"topic" toml:"topic"`
+	PublishTopic        string     `mapstructure:"publish_topic" toml:"publish_topic"`
+	Group               string     `mapstructure:"group" toml:"group"`
+	Enabled             bool       `mapstructure:"enabled" toml:"enabled"`
+	OrgWhitelist        mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
+	OrgWhitelistEnabled bool       `mapstructure:"enable_org_whitelist" toml:"enable_org_whitelist"`
```
Sergey1011010, line 33: I'd keep field name and tag consistent, but it can be a nitpick :)

configuration.go
```diff
@@ -50,7 +50,7 @@ var config struct {
 	Broker     broker.Configuration `mapstructure:"broker" toml:"broker"`
 	Server     server.Configuration `mapstructure:"server" toml:"server"`
 	Processing struct {
-		OrgWhiteListFile string `mapstructure:"org_white_list_file" toml:"org_white_list_file"`
+		OrgWhiteListFile string `mapstructure:"org_whitelist" toml:"org_whitelist"`
```
Sergey1011010, line 53: maybe it's better to keep it with `_file` since it's a path to the file?

Sergey1011010:
Please, update the last sentence here https://github.com/RedHatInsights/insights-results-aggregator#whole-data-flow

> Please note that results are filtered - only results for organizations listed in org_whitelist.csv are processed and cached in aggregator.

quarckster:
Please don't forget to update the config in e2e-deploy.

configuration.go
JoseLSegura: +1

broker/configuration.go
bond95, line 33: Maybe instead of adding new flag, will be better to set `org_white_list` as empty string to indicate whitelist disabling?

tisnik: dunno, would be nice to be able to distinguish between disabled whitelist and missing environment variable (and we forgot about it several times :-)

configuration.go
Bee-lee: @Sergey1011010 @JoseLSegura this was actually a bug, the problem with this is #577. It was never loaded from the path in .tomls. 
I'll change the name of the variable in the configs, it makes sense

broker/configuration.go
Bee-lee: @bond95 I was thinking the same thing, but later went with this because of @tisnik reasons.

Bee-lee: fixd, thanks

broker/configuration.go
```diff
@@ -24,11 +24,12 @@ import (
...
-	Group        string     `mapstructure:"group" toml:"group"`
-	Enabled      bool       `mapstructure:"enabled" toml:"enabled"`
-	OrgWhitelist mapset.Set `mapstructure:"org_white_list" toml:"org_white_list"`
-	SaveOffset   bool       `mapstructure:"save_offset" toml:"save_offset"`
+	Address             string     `mapstructure:"address" toml:"address"`
+	Topic               string     `mapstructure:"topic" toml:"topic"`
+	PublishTopic        string     `mapstructure:"publish_topic" toml:"publish_topic"`
+	Group               string     `mapstructure:"group" toml:"group"`
+	Enabled             bool       `mapstructure:"enabled" toml:"enabled"`
+	OrgWhitelist        mapset.Set `mapstructure:"org_whitelist" toml:"org_whitelist"`
```
quarckster, line 32: Why do you make `toml:"org_whitelist"`` but in the config you use `org_whitelist_file`?

Bee-lee: @quarckster thanks, last place I missed....

Bee-lee: @quarckster  this is actually not wrong, because this is broker/ configuration and the OrgWhitelist is not loaded automatically, but by calling `config.Broker.OrgWhitelist = getOrganizationWhitelist()`, so it will still work

but I'm fixing it

configuration.go
Bee-lee: done

broker/configuration.go
Bee-lee: done

PR #639, opened 2020-04-24
add ability to disable/enable a rule for specific cluster

### LINKED ISSUE #302: Ability to disable a rule for selected cluster - backend part
UI part: #169 

The user wants to disable rules that they're not interested in for a specific cluster in order to hide them from the report. 

Therefore, the user should also have the ability to:
- see the rules that have been disabled (+possibly see when that happened)
- re-enable them (+possibly see when that happened)

## Tasks:
- [ ] create storage part for cluster+organization+rule
    - schema should include date of disablement and possibly date of re-enablement

- [ ] create endpoints for disabling and enabling rules per cluster and organization.
    - allow multiple rules at once, so the user would be able to possibly disable a filtered selection of rules

- [ ] decide on the method of serving the rules, depending on UI decision, and implement it. see below. I'll update it later.

### Possible implementations of serving the disabled rules

1) API endpoint for report `report/{organization}/{cluster}` always returns all rules - separated into two lists of hit rules (as present) and disabled rules. 
    - this would be better if the disabled rules were in the same tab/component, without having to reload it. if the request for disable/enable would return an error, frontend would handle it with a callback.
2) API endpoint for report returns only enabled rules. Disabled rules are served as a separate endpoint, probably `clusters/{cluster}/rules`
    - this would be better if the disabled rules were in a different tab/component than the normal rules/report which would require a tab/table switch anyway

Bee-lee:
# Description
This adds the ability to disable a rule for specified cluster. User also has the ability to re-enable the rule.

As per the mockups, the disabled rules are still returned on the report endpoint, but now they have another attribute `"disabled": true/false`. @bond95 @Sergey1011010 Is this OK for you? The default database sorting is now `enabled` rules first. 

I also propose changes to the naming of migrations, so that they would indicate what's in them just by the name. @natiiix and everyone, tell me what you think about that.

Example response:
```
      {
        "rule_id": "ccx_rules_ocm.tutorial_rule",
        "description": "Introducing Insights for Red Hat OpenShift Container Platform",
        "details": "Red Hat Insights for OpenShift is a proactive management solution. It provides ongoing infrastructure analyses of your Red Hat OpenShift Container Platform 4.2 and later installations. Red Hat Insights helps you identify, prioritize, and resolve risks to security, performance, availability, and stability before they become urgent issues.\nRed Hat Insights for OpenShift uses the Remote Health Monitoring feature of OpenShift 4. The health checks are created by Red Hat subject matter experts and assessed according to severity and impact.\nThis is an example  recommendation that you can safely ignore. To disable it, click  the triple-dot menu button next to the header, and select Disable.\n",
        "created_at": "2020-04-08T00:42:00Z",
        "total_risk": 1,
        "risk_of_change": 0,
        "extra_data": {
          "error_key": "TUTORIAL_ERROR",
          "type": "rule"
        },
        "disabled": false
      },
```

Fixes #302 

## Type of change

- Documentation update

## Testing steps
`make before_commit` seems to always time out on postgresql for some reason, but it passed CI


migration/migration.go
```diff
@@ -43,11 +43,12 @@ type Migration struct {
...
-	mig2,
-	mig3,
-	mig4,
-	mig5,
+	mig0001CreateReport,
+	mig0002CreateRuleContent,
+	mig0003CreateClusterRuleUserFeedback,
+	mig0004ModifyClusterRuleUserFeedback,
+	mig0005CreateConsumerError,
+	mig0006CreateClusterRuleToggle,
```
Sergey1011010, line 51: Good idea!

server/router_utils.go
```diff
@@ -236,46 +236,40 @@ func readErrorKey(writer http.ResponseWriter, request *http.Request) (types.Erro
...
-func (server *HTTPServer) readVoteOnRuleParams(
+// readClusterRuleUserParams gets cluster_name, rule_id and user_id from current request
+func (server *HTTPServer) readClusterRuleUserParams(
 	writer http.ResponseWriter, request *http.Request,
-) (types.ClusterName, types.RuleID, types.UserID, bool) {
+) (types.ClusterName, types.RuleID, types.UserID, error) {
 	clusterID, err := readClusterName(writer, request)
 	if err != nil {
 		// everything has been handled already
-		return "", "", "", false
```
Sergey1011010, line 246: why to return error when we're not using it anywhere?

Bee-lee: At first, I thought that the `bool` here is mainly for the `checkVotePermission` which I split, so I changed this function to be consistent with the rest of the functions in `router_utils`, they always return the errors even though everything has been handled already (logged + response sent) @Sergey1011010

migration/migration.go
Bee-lee: thanks:)

storage/storage_rules_test.go
```diff
@@ -490,6 +495,110 @@ func TestDBStorageGetContentForRulesRowsError(t *testing.T) {
...
+		mustWriteReport3Rules(t, mockStorage)
+
+		helpers.FailOnError(t, mockStorage.ToggleRuleForCluster(
+			testdata.ClusterName, testdata.Rule1ID, testdata.UserID, state,
+		))
+
+		_, err := mockStorage.GetFromClusterRuleToggle(testdata.ClusterName, testdata.Rule1ID, testdata.UserID)
+		helpers.FailOnError(t, err)
+
+		helpers.FailOnError(t, mockStorage.Close())
```
Sergey1011010, line 514: you don't have to call close since closer already does that

Sergey1011010: since it's for loop where you create storage per test I'd wrap it to a function

server/server.go
```diff
@@ -541,6 +658,9 @@ func (server *HTTPServer) addEndpointsToRouter(router *mux.Router) {
 	router.HandleFunc(apiPrefix+DislikeRuleEndpoint, server.dislikeRule).Methods(http.MethodPut, http.MethodOptions)
 	router.HandleFunc(apiPrefix+ResetVoteOnRuleEndpoint, server.resetVoteOnRule).Methods(http.MethodPut, http.MethodOptions)
 	router.HandleFunc(apiPrefix+ClustersForOrganizationEndpoint, server.listOfClustersForOrganization).Methods(http.MethodGet)
+	router.HandleFunc(apiPrefix+DisableRuleForClusterEndpoint, server.disableRuleForCluster).Methods(http.MethodPut, http.MethodOptions)
+	router.HandleFunc(apiPrefix+EnableRuleForClusterEndpoint, server.enableRuleForCluster).Methods(http.MethodPut, http.MethodOptions)
+	router.HandleFunc(apiPrefix+ListDisabledRulesForClusterEndpoint, server.listDisabledRulesForCluster).Methods(http.MethodGet)
```
bond95, line 663: Please, instead of using direct endpoint to disabled rules, better add query `disabled` to `ReportEndpoint`. It will simplify filtering. Without `disabled` query let it return all rules.

Sergey1011010: Agree. 

And also according to the mock https://marvelapp.com/852jaj9/screen/68494249 it looks like we would need a way to fetch 1 all rules, 2 disabled rules and 3 enabled rules. But also all of this could be done on frontend the same way as we did for another filters.

Sergey1011010:
Is `disabled` field returned now by report related endpoints? Would be great to have it, because filters can be set to show both.

server/server.go
Bee-lee: Are you guys referring the List endpoint? I was using it a lot when I was testing and I left it included in the final endoints.go by mistake (I'll remove it). I didn't want to delete it completely because I think it can be used for some tests.

Regarding the filtering - I assumed that this would be done on frontend like all the other filtering and sorting and considering that the design/mockup is missing information about what is default (I guess Enabled?), and since the "Status" is just yet another filter, I don't think it wouldn't make sense for it to behave differently from other filters - reload (and wait for) the whole results just for one keyword. 

@bond95 @Sergey1011010

Sergey1011010: @Bee-lee then this endpoint should be included only in debug mode. Thanks for clarification, you're right, filtering can be done on frontend.

bond95: > Are you guys referring the List endpoint? I was using it a lot when I was testing and I left it included in the final endoints.go by mistake (I'll remove it). I didn't want to delete it completely because I think it can be used for some tests.


@Bee-lee yes, endpoint with query `disabled` could be used for tests same as this endpoint, but without additional endpoint.

> Regarding the filtering - I assumed that this would be done on frontend like all the other filtering and sorting and considering that the design/mockup is missing information about what is default (I guess Enabled?), and since the "Status" is just yet another filter, I don't think it wouldn't make sense for it to behave differently from other filters - reload (and wait for) the whole results just for one keyword.

Yes, filtering can be done on frontend, but I still don't see any reason to use separate endpoint. I prefer to have endpoint which can return enabled or disabled or both rules according to query, because it's more flexible.

server/router_utils.go
Bee-lee: @Sergey1011010 I agree that it doesnt make any sense to return errors when they are already logged and handled. We can refactor this later and change all these "everything has been handled" errors to "ok" booleans, but for me, consistency is key

storage/storage_rules_test.go
Bee-lee: right, I see, thanks :+1: fixd :)

PR #690, opened 2020-05-04
Fix in script check_coverage.sh

### LINKED ISSUE #689: Don't use hardcoded terminal colors in `check_coverage.sh`
Currently, hardcoded terminal colors are used in `check_coverage.sh`. It would be better to use tput command instead. Additionaly prob. OK/Fail message will be more visible by using background colors.

tisnik:
# Description

Fixes #689 

## Type of change

### DIFF
--- a/check_coverage.sh
+++ b/check_coverage.sh
@@ -15,9 +15,10 @@
 
 
 THRESHOLD=85
-RED='\033[0;31m'
-GREEN='\033[0;32m'
-NC='\033[0m' # No Color
+
+RED_BG=$(tput setab 1)
+GREEN_BG=$(tput setab 2)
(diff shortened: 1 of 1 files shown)


check_coverage.sh
```diff
@@ -15,9 +15,10 @@
 
 
 THRESHOLD=85
-RED='\033[0;31m'
-GREEN='\033[0;32m'
-NC='\033[0m' # No Color
+
+RED_BG=$(tput setab 1)
+GREEN_BG=$(tput setab 2)
+NC=$(tput sgr0) # No Color
```
JoseLSegura, line 21: I haven't seen this way to define the colors. Nice to learn :-D

PR #694, opened 2020-05-04
Skip outdated reports in consumer (using LastChecked cache instead of an extra SELECT query)

### LINKED ISSUE #664: Cache for 'LastChecked' attribute for incoming messages and the ability to drop older reports
It might be good to have a cache for 'LastChecked' attribute for incoming messages. It allow us to quickly drop older reports w/o the need to access DB which is bottleneck as was measured there:
https://github.com/tisnik/jupyter-notebook-examples/blob/master/consumer_benchmarks.ipynb

Algorithm:

```
type ClusterName string
type LastChecked time.Time

cache := map[clusterName]lastChecked

func Consumer(msg Message) {
     stored := cache[msg.ClusterName]
     if stored == nil || msg.lastChecked > cache[msg.ClusterName] {
         db.Store(msg)
         cache[msg.ClusterName] = msg.lastChecked
    } else {
         log.Info("Message has been dropped")
    }
}
```

Acceptance criteria:
- this needs to be configurable

natiiix:
# Description

Right now, every time a report is about to be written into database, we first check if it isn't older than an existing report already in the database. This approach is very slow because it required the execution of SQL query for every outdated report.

To make this check significantly faster, the `last_checked_at` values will be loaded into an in-memory dictionary and no SQL transaction will begin if the report is outdated. In case there will be multiple consumer instances running in parallel, the old check still remains present. However, it won't be executed for obviously/significantly outdated reports, which should still speed up the initial consumption of the already processed reports that happens due to some issues with the Kafka library.

Fixes #664 

## Type of change

## Testing steps

`make before_commit` as usual, including postgres-based tests

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -205,7 +205,9 @@ func (storage DBStorage) Init() error {
 		)
 
 		if err := rows.Scan(&clusterName, &lastChecked); err != nil {
-			_ = rows.Close()
+			if closeErr := rows.Close(); closeErr != nil {
+				log.Error().Err(closeErr).Msg("Unable to close the DB rows handle")
+			}
 			return err
 		}
 
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -99,7 +99,7 @@ type DBDriver int
 const (
 	// DBDriverSQLite3 shows that db driver is sqlite
 	DBDriverSQLite3 DBDriver = iota
-	// DBDriverPostgres shows that db driver is postrgres
+	// DBDriverPostgres shows that db driver is postgres
 	DBDriverPostgres
 	// DBDriverGeneral general sql(used for mock now)
 	DBDriverGeneral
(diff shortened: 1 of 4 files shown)


storage/storage.go
```diff
@@ -181,11 +184,38 @@ func initAndGetDriver(configuration Configuration) (driverType DBDriver, driverN
...
+
+	for rows.Next() {
+		var (
+			clusterName types.ClusterName
+			lastChecked time.Time
+		)
 
-	return migration.SetDBVersion(storage.connection, migration.GetMaxVersion())
+		if err := rows.Scan(&clusterName, &lastChecked); err != nil {
+			_ = rows.Close()
```
Sergey1011010, line 208: maybe would be better to log an error here

natiiix: Ok, is it better like this?

natiiix:
> look good. Btw it could be improved by adding `where` condition after do update set https://www.postgresql.org/docs/9.5/sql-insert.html because it's still possible to write old result when we have multiple pods and one of the pods have outdated cache.

@Sergey1011010 It's possible, but quite unlikely since there's still that `SELECT`-based check and that's inside of a transaction, so we should never overwrite a newer report with an older one.

PR #709, opened 2020-05-06
Building images changed registry.redhat.io/rhel8/go-toolset

### LINKED ISSUE #90: Story: Auth. scheme for accessing Insigts results from OCM

### LINKED ISSUE #708: Use Red Hat image for building
In order to comply with security policies, we should use only Red Hat images.

quarckster:
# Description
We should use only Red Hat images.

Fixes #708 

## Type of change

## Testing steps

Image has been built locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,35 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13 AS builder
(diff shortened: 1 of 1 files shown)


quarckster:
Shame on me :)

PR #716, opened 2020-05-06
Fixing Makefile to execute shellcheck over the whole Bash base code

### LINKED ISSUE #715: Shellcheck is not being executed for all Bash scripts
I just realized that `shellcheck` rule in `Makefile` only run the checks for `local_storage/populate_db_with_mock_data.sh` and `utils/fill_in_results.sh`.

In fact, it is hiding some shell warnings in several scripts in the root directory

JoseLSegura:
# Description

The current `make shellcheck` execution only checks a minimal subset of our Bash scripts. It is needed to run it over the whole Bash codebase.

Fixes #715

## Type of change

## Testing steps

CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help clean build fmt lint vet run test cover style cyclo
+.PHONY: default clean build fmt lint vet cyclo ineffassign shellcheck errcheck goconst gosec abcgo json-check style run test cover integration_tests rest_api_tests rules_content sqlite_db license before_commit help
 
 SOURCES:=$(shell find . -name '*.go')
 
@@ -31,7 +31,7 @@ ineffassign: ## Run ineffassign checker
 	./ineffassign.sh
 
 shellcheck: ## Run shellcheck
-	shellcheck **/*.sh
+	shellcheck *.sh */*.sh
 
 errcheck: ## Run errcheck
 	@echo "Running errcheck"
--- a/gosec.sh
+++ b/gosec.sh
@@ -20,11 +20,6 @@ NC=$(tput sgr0) # No Color
 
 GO_SEC_ARGS=""
 
-VERBOSE=false
-if [[ $* == *verbose* ]]; then
-    VERBOSE=true
-fi
-
 if [[ $* != *verbose* ]]; then
     GO_SEC_ARGS="-quiet"
 fi
--- a/test.sh
+++ b/test.sh
@@ -95,8 +95,9 @@ function test_rest_api() {
     # Without this, the DB could be locked because
     # the migrations have not yet finished.
     sleep 2
-    for i in {1..5}; do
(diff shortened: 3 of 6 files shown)


gosec.sh
```diff
@@ -20,11 +20,6 @@ NC=$(tput sgr0) # No Color
 
 GO_SEC_ARGS=""
 
-VERBOSE=false
-if [[ $* == *verbose* ]]; then
-    VERBOSE=true
-fi
-
```
JoseLSegura, line 27: Shellcheck reported that `VERBOSE` is not used, but I don't know if `go sec` use it behind the scenes. Could you confirm?

test.sh
```diff
@@ -95,8 +95,9 @@ function test_rest_api() {
     # Without this, the DB could be locked because
     # the migrations have not yet finished.
     sleep 2
-    for i in {1..5}; do
-        populate_db_with_mock_data 2>/dev/null && break || sleep 1
+    for _ in {1..5}; do
+        populate_db_with_mock_data 2>/dev/null && break
+        sleep 1
```
JoseLSegura, line 100: I guess that the meaning of both versions is equivalent, but the previous one reported [`SC2015: Note that A && B || C is not if-then-else. C may run when A is true.`](https://www.shellcheck.net/wiki/SC2015)

unit-tests.sh
```diff
@@ -15,11 +15,12 @@
 
 
 CONFIG_FILE=config-devel.toml
-POSTGRES_PORT=$(cat $CONFIG_FILE | grep -i pg_port | grep -E -o '[0-9]+')
+POSTGRES_PORT=$(grep -i pg_port $CONFIG_FILE | grep -E -o '[0-9]+')
 
 function run_unit_tests() {
-    go test -coverprofile coverage.out $(go list ./... | grep -v tests)
-    if [ $? -ne 0 ]; then
+    # shellcheck disable=SC2046
```
JoseLSegura, line 21: I was totally unable to resolve this problem without breaking tests execution, so I disabled this warning

Makefile
```diff
@@ -31,7 +31,7 @@ ineffassign: ## Run ineffassign checker
 	./ineffassign.sh
 
 shellcheck: ## Run shellcheck
-	shellcheck **/*.sh
+	shellcheck *.sh */*.sh
```
Sergey1011010, line 34: Probably `find . -exec` would also work here

gosec.sh
Sergey1011010: the next `if` was supposed to be not the same as this one, but to check `VERBOSE` variable

Makefile
JoseLSegura: Maybe, but we can have Python scripts that are executable too, so I prefer to find them using the extension (probably using `find . -name "*.sh"` could be better...

PR #717, opened 2020-05-06
Add minimal migration CLI functionality

### LINKED ISSUE #541: Migrations CLI
Migrations should be executed via CLI. The intention is performance optimizations. Every time when the executable is started it performs migrations. It can affect the starting time of pods in OCP and the performance of horizontal pod autoscaler. We should perform migrations only once per deployment using pod-based lifecycle hook https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/deployment_strategies.html#pod-based-lifecycle-hook.

Acceptance criteria:
* migrations can be executed via cli, e.g.: `insights-results-aggregator migrations upgrade latest|migration_number`, `insights-results-aggregator migrations downgrade base|migration_number`.

natiiix:
# Description

Not sure if we're really gonna use this, but I believe it could do the job, at least until we decide to move onto a real migration library/framework.

- [x] Executing the `insights-results-aggregator migration latest` command will update the database to the latest migration version.
- [x] Regular start-up of the Aggregator (in service mode) will not perform any DB migrations.

*Marked as WIP because it's WIP, but also because currently the fuction has a ridiculously high cyclomatic complexity, so it may be a bit difficult to review with confidence.*

Fixes #541

## Type of change

## Testing steps

Just manually testing the command so far.

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -337,7 +337,7 @@ func printMigrationInfo(dbConn *sql.DB) int {
 	currMigVer, err := migration.GetDBVersion(dbConn)
 	if err != nil {
 		log.Error().Err(err).Msg("Unable to get current DB version")
-		return ExitStatusPrepareDbError
+		return ExitStatusMigrationError
 	}
 
 	log.Info().Msgf("Current DB version: %d", currMigVer)
@@ -354,7 +354,7 @@ func setMigrationVersion(dbConn *sql.DB, versStr string) int {
 		vers, err := strconv.Atoi(versStr)
 		if err != nil {
 			log.Error().Err(err).Msg("Unable to parse target migration version")
-			return ExitStatusPrepareDbError
+			return ExitStatusMigrationError
 		}
 
 		targetVersion = migration.Version(vers)
@@ -362,7 +362,7 @@ func setMigrationVersion(dbConn *sql.DB, versStr string) int {
 
 	if err := migration.SetDBVersion(dbConn, targetVersion); err != nil {
 		log.Error().Err(err).Msg("Unable to perform migration")
-		return ExitStatusPrepareDbError
+		return ExitStatusMigrationError
 	}
 
 	log.Info().Msgf("Database version is now %d", targetVersion)
@@ -389,7 +389,7 @@ func performMigrations() int {
 
 	default:
 		log.Error().Msg("Unexpected number of arguments to migrations command (expected 0-1)")
-		return ExitStatusPrepareDbError
+		return ExitStatusMigrationError
 	}
 }
 
--- a/aggregator.go
+++ b/aggregator.go
@@ -77,6 +77,11 @@ var (
 
 	// BuildCommit contains Git commit used to build this application
 	BuildCommit string = "*not set*"
+
+	// autoMigrate determines if the prepareDB function upgrades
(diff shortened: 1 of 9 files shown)


Bee-lee:
I like this!
Do you think adding the option to specify the DB could be useful for us?
e.g. `insights-results-aggregator migrate ./test.db latest` or something

natiiix:
> I like this!
> Do you think adding the option to specify the DB could be useful for us?
> e.g. `insights-results-aggregator migrate ./test.db latest` or something

@Bee-lee It uses the DB from config.toml, so you can set it there. I don't see that much of a reason to allow any complex configuration like this because SQLite isn't used in production and I doubt anyone would want to specify the connection path to a custom Postgres DB.

Bee-lee:
@natiiix right, of course... So I guess `INSIGHTS_RESULTS_AGGREGATOR__STORAGE__SQLITE_DATASOURCE` would work too, thanks:)

natiiix:
> @natiiix right, of course... So I guess `INSIGHTS_RESULTS_AGGREGATOR__STORAGE__SQLITE_DATASOURCE` would work too? thanks:)

@Bee-lee I don't really know, but anything that will be processed when reading the storage configuration should work. I believe the code is shared by all part of the service.

aggregator.go
```diff
@@ -305,6 +307,63 @@ func printEnv() int {
...
+	if err != nil {
+		log.Error().Err(err).Msg("Unable to prepare DB for migrations")
+		return ExitStatusPrepareDbError
+	}
+	defer closeStorage(db)
+
+	dbConn := db.GetConnection()
+
+	if err := migration.InitInfoTable(dbConn); err != nil {
+		log.Error().Err(err).Msg("Unable to initialize migration info table")
```
Sergey1011010, line 323: forgot return?

natiiix: @Sergey1011010 Yes, thanks, I knew I definitely missed one of those on-error returns.

aggregator_test.go
```diff
@@ -70,6 +70,10 @@ func TestCreateStorage(t *testing.T) {
 }
 
 func TestStartService(t *testing.T) {
+	// It is necessary to perform migrations for this test
+	// because the service won't run on top of an empty DB.
+	*main.AutoMigratePtr = true
```
Sergey1011010, line 75: Maybe something like `EnableAutoMigrate()` and `defer DisableAutoMigrate()` would be better, but it's definitely a nitpick :)

test.sh
```diff
@@ -15,7 +15,7 @@
 
 COLORS_RED='\033[0;31m'
 COLORS_RESET='\033[0m'
-LOG_LEVEL="fatal"
+LOG_LEVEL="error"
```
Sergey1011010: it makes test.sh loud af, cuz there's lots of tests which are expected to fail, I'd just add `--verbose` flag to the CI.

aggregator_test.go
natiiix: @Sergey1011010 I wanted to avoid adding too much code for this because I don't think it should be used from too many places.

test.sh
natiiix: This was just a temporary change, so that I could see what was wrong because it was impossible to determine the origin of the issue from the generic "service exited with error" message.

PR #721, opened 2020-05-07
Colorized json checker output

tisnik:
# Description

Colorized JSON checker output.

## Type of change

## Howto test

See it in action https://travis-ci.org/github/RedHatInsights/insights-results-aggregator/jobs/684139931


tisnik:
> I know this is "only" a tool, but why not using `colorama` (if available)? Creation a subprocess only for getting the color codes looks a bit overkiller to me :-D

it is not available on CI, and I don't want to touch Makefile much (it will be larger than these four lines :)

PR #740, opened 2020-05-12
Print error on outdated DB migration version

natiiix:
# Description

If the DB version is not the latest, the aggregator service will not start.

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -105,6 +105,30 @@ func closeStorage(storage *storage.DBStorage) {
 	}
 }
 
+func prepareDBMigrations(dbStorage *storage.DBStorage) int {
+	// This is only used by some unit tests.
+	if autoMigrate {
+		if err := dbStorage.MigrateToLatest(); err != nil {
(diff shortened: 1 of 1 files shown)


aggregator.go
```diff
@@ -105,6 +105,30 @@ func closeStorage(storage *storage.DBStorage) {
 	}
 }
 
+func prepareDBMigrations(dbStorage *storage.DBStorage) int {
+	// This is only used by some unit tests.
+	if autoMigrate {
```
tisnik, line 110: do we need to use global var there?

natiiix: @tisnik The only alternative I could think of was adding a field into the config file, but that felt like even more effort.

tisnik: I meant that `autoMigrate` might be parameter passed into the function.

tisnik: I don't want to stop your work and this one is very minor thing

PR #750, opened 2020-05-13
Fixed the offset problem

### LINKED ISSUE #400: Add support of different migrations for different DBs

### LINKED ISSUE #657: Fix for the problems with not stored offsets in Kafka broker
### 1. Use some in memory storage which caches to disk(like Redis)

- Pros:
    - Fast as fuck

- Cons:
    - Extra deployment
    - Extra point of failure

### 2. Store in each record of report table

- Pros:
    - Ability to track last message offset for each report(do we need it?)
- Cons:
    - To be able to get the latest offset we would need an index which brings little overhead to insert

### 3. Store in another table as single row
- Pros:
    - No that much overhead as having index
- Cons:
    - Still not fast as fuck because we’re doing another query, but totally fine since we’re not doing “Big Data”

### 4. Make kafka library work by either figuring out why it throws weird errors or using another library. 
- Pros:
    - not much work probably
- Cons:
    - would be nice to know the offset, at least for debugging purposes

### 5. Keep everything as is and suffer
- Pros:
    - already done :)
- Cons: see description

### LINKED ISSUE #751: Make migrations tests work with postgres

Sergey1011010:
# Description

Tried to split to different commits so that it's not that overwhelming and refactoring doesn't distract your attention from actual code. 

Fixes #657
Fixes #400
Fixes #751

## Type of change

## Testing steps

`make before_commit`


types/errors.go
```diff
@@ -0,0 +1,178 @@
...
+
+func convertPostgresError(err error) error {
+	pqError, ok := err.(*pq.Error)
+	if !ok {
+		return err
+	}
+
+	// see https://www.postgresql.org/docs/current/errcodes-appendix.html to get the magic happening below
+	switch pqError.Code {
+	case "42P07": // duplicate_table
```
tisnik, line 133: prob. worth to use `const` there in the future. But let's not delay this fix to be deployed.

PR #756, opened 2020-05-14
Update stable

quarckster:
I tested fixes in
https://github.com/RedHatInsights/insights-results-aggregator/pull/750
https://github.com/RedHatInsights/insights-results-aggregator/pull/753
in CI env and offset is stored in the db as expected.


Bee-lee:
lets go!

PR #775, opened 2020-05-15
Fix commiting offset to kafka

Sergey1011010:
# Description

Rewrote consumer to use groups properly. Now it commits messages and start from the latest offset by magic. 

Fixes #695
Should fix #620

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update
- Refactor (refactoring code, removing useless files)
- Unit tests (no changes in the code)

## Testing steps

`make before_commit` & test that messages are commited in dev pod

PR #776, opened 2020-05-15
Dirty integration of Payload Tracker service

### LINKED ISSUE #622: Payload tracker integration
In order to track a payload in the whole pipeline starting with ingress service, we should integrate with payload tracker service.

Ingress service return so-called request id. Having this request id you can track when your payload is handled by various services. At the moment the latest service where we can track insights operator archives is insights-storage-broker.

Here is the documentation https://github.com/RedHatInsights/payload-tracker#integration

Acceptance criteria:

 *  I should be able to see insights-results-aggregator in payload tracker when I enter request id returned by ingress service.

natiiix:
# Description

**TBD**

Fixes #622 

## Type of change

## Testing steps

`make before_commit` & manual check of the Payload Tracker topic in Kafka using `kafkacat`

### DIFF
--- a/config-devel.toml
+++ b/config-devel.toml
@@ -1,7 +1,8 @@
 [broker]
 address = "kafka:29092"
 topic = "ccx.ocp.results"
-publish_topic = "platform.payload-status"
+payload_tracker_topic = "platform.payload-status"
+service_name = "insights-results-aggregator"
 group = "aggregator"
 enabled = true
 enable_org_whitelist = false
--- a/config-devel.toml
+++ b/config-devel.toml
@@ -1,6 +1,7 @@
 [broker]
 address = "kafka:29092"
 topic = "ccx.ocp.results"
+publish_topic = "platform.payload-status"
 group = "aggregator"
 enabled = true
 enable_org_whitelist = false
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -20,7 +20,6 @@ package consumer
 
 import (
 	"context"
-	"time"
 
 	"github.com/Shopify/sarama"
 	"github.com/rs/zerolog/log"
@@ -35,7 +34,7 @@ import (
 type Consumer interface {
 	Serve()
 	Close() error
-	ProcessMessage(msg *sarama.ConsumerMessage) (MaybeRequestID, error)
+	ProcessMessage(msg *sarama.ConsumerMessage) (types.RequestID, error)
 }
 
 // KafkaConsumer in an implementation of Consumer interface
@@ -89,7 +88,7 @@ func NewWithSaramaConfig(
 		return nil, err
 	}
 
-	ptProducer, err := producer.New(brokerCfg)
+	payloadTrackerProducer, err := producer.New(brokerCfg)
 	if err != nil {
 		log.Error().Err(err).Msg("unable to construct producer")
 		return nil, err
@@ -102,7 +101,7 @@ func NewWithSaramaConfig(
 		numberOfSuccessfullyConsumedMessages: 0,
 		numberOfErrorsConsumingMessages:      0,
 		ready:                                make(chan bool),
-		payloadTrackerProducer:               ptProducer,
+		payloadTrackerProducer:               payloadTrackerProducer,
 	}
 
 	return consumer, nil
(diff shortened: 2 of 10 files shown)


consumer/consumer.go
```diff
@@ -54,7 +55,7 @@ const (
 type Consumer interface {
 	Serve()
 	Close() error
-	ProcessMessage(msg *sarama.ConsumerMessage) error
+	ProcessMessage(msg *sarama.ConsumerMessage) (interface{}, error)
```
tisnik, line 58: that's pretty untyped :)

consumer/consumer.go
```diff
@@ -92,24 +95,33 @@ func New(brokerCfg broker.Configuration, storage storage.Storage) (*KafkaConsume
...
-	brokerCfg broker.Configuration,
-	storage storage.Storage,
-	saramaConfig *sarama.Config,
-	saveOffset bool,
-) (*KafkaConsumer, error) {
+func createKafkaConsumer(brokerCfg broker.Configuration, saramaConfig *sarama.Config) (sarama.Client, sarama.Consumer, []int32, error) {
 	client, err := sarama.NewClient([]string{brokerCfg.Address}, saramaConfig)
 	if err != nil {
-		return nil, err
+		return nil, nil, nil, err
```
tisnik, line 101: this reminds me code (constructors) in ccx-data-pipeline :)

consumer/consumer.go
```diff
@@ -258,6 +277,41 @@ func organizationAllowed(consumer *KafkaConsumer, orgID types.OrgID) bool {
...
 
+func (consumer *KafkaConsumer) trackPayload(reqID interface{}, timestamp time.Time, status string) {
+	reqIDString, ok := reqID.(string)
+	if !ok {
+		log.Warn().Msgf("request ID is not a string: %#v", reqID)
+		return
+	}
+
+	_, _, err := consumer.payloadTrackerProducer.ProduceMessage(producer.PayloadTrackerMessage{
+		Service:   "insights-results-aggregator",
```
tisnik, line 288: ok for PoC, might be better to have it configurable later

Sergey1011010:
Please, reassign the issue in github and jira then.

Sergey1011010:
Also take a look at https://github.com/RedHatInsights/insights-results-aggregator/pull/775 cuz a lot of things has been changed in consumer.

consumer/processing.go
```diff
@@ -30,13 +30,19 @@ import (
 // Report represents report send in a message consumed from any broker
 type Report map[string]*json.RawMessage
 
+// MaybeRequestID contains either a string or nil depending on the content
+// of the input message. Sometimes the request ID is missing,
+// in which case it is nil, otherwise it should always be a string.
+type MaybeRequestID interface{}
```
Sergey1011010, line 36: What about to use a pointer to a string or a struct with field `IsNull`?

Sergey1011010:
Please, update producer's documentation since now it's producing only payload tracker messages

config-devel.toml
```diff
@@ -1,6 +1,7 @@
 [broker]
 address = "kafka:29092"
 topic = "ccx.ocp.results"
+publish_topic = "platform.payload-status"
```
JoseLSegura, line 4: I know that this is a WIP, so sorry if it is a nitpick.

In theory, the Kafka server can be different than the consumer one, so maybe it deserves a new section in the configuration file for the payload tracker integration where the address and topic are configured

natiiix: That makes sense, I guess. It is getting a bit crowded in the `broker` section. I'd rather add these tweaks in a separate PR though because this one already requires a lot of changes to many different files, so the longer I keep it open, the more conflicts I have to resolve every day.

consumer/processing.go
natiiix: In the end, it runs out that using a regular `string` works just fine.

consumer/consumer.go
natiiix: Now configurable via `broker.service_name`.

consumer/consumer.go
natiiix: Haha, yeah, I thought of that too when I wrote it. I believe these return of many `nil`s is no longer in the refactored code now, thankfully.

consumer/consumer.go
natiiix: Didn't realize it was possible to use `string` when unmarshaling `null` and missing fields, but apparently it's fine, so we no longer have to use this ugly trick.

config-devel.toml
```diff
@@ -1,6 +1,8 @@
 [broker]
 address = "kafka:29092"
 topic = "ccx.ocp.results"
+payload_tracker_topic = "platform.payload-status"
```
quarckster, line 4: Please don't forget to update respectful config in e2e-deploy repo.

PR #795, opened 2020-05-20
Save tags to db

### LINKED ISSUE #790: Add mechanism to save and retrieve rule tags

Bee-lee:
# Description
This adds a (non-optimal) solution to saving rule tags into the database and retrieving them for the API.

Fixes #790 

## Type of change

## Testing steps
`make before_commit` finished with some weird timeout, let's see here


Sergey1011010:
I think it would be better to use array type, but it's too late :)

PR #809, opened 2020-05-22
Update stable

quarckster:


quarckster:
There is a problem with smoke tests.

JoseLSegura:
@quarckster apparently the smoke test problem is related to not being able to download Molodec from Gitlab... I don't know if Jenkins machines doesn't have access to it or it is some punctual problem with resolving the DNS name

PR #812, opened 2020-05-25
Added jenkins config to run unit tests in OCP

quarckster:
# Description

This PR moves all checks from Travis to Jenkins. Please pay attention to `test` and `test-postgres` targets. `before_commit` target includes only `test`. In Jenkins only `test-postgres` is executed.

## Type of change


config-devel.toml
```diff
@@ -26,7 +26,7 @@ db_driver = "postgres"
 pg_username = "user"
 pg_password = "password"
 pg_host = "localhost"
-pg_port = 55432
+pg_port = 5432
```
Sergey1011010, line 29: Why did you change a port here?

quarckster: Because 5432 it's a default port for https://catalog.redhat.com/software/containers/detail/5aa63541ac3db95f196086f1?container-tabs=dockerfile. I assume this config is used for unit and integration tests.

Sergey1011010: Then we would need to update docker-compose.yml for local development as well, not a big deal tho

Jenkinsfile
```diff
@@ -0,0 +1,36 @@
...
+
+        gitUtils.stageWithContext("Integration-tests", shortenURL = false) {
+            unitTestsStatus = sh(script: "make integration_tests", returnStatus: true)
+        }
+
+        if (unitTestsStatus != 0) {
+            error("Integration tests failed")
+        }
+    }
+}
```
natiiix, line 36: Missing final EOL, is this intentional?

unit-tests.sh
```diff
@@ -26,16 +24,16 @@ function run_unit_tests() {
...
+if [ "$STORAGE" == "postgres" ]; then
     # tests with postgres
-    export INSIGHTS_RESULTS_AGGREGATOR__TESTS_DB="aggregator"
-    export INSIGHTS_RESULTS_AGGREGATOR__TESTS_DB_ADMIN_PASS="password"
-    echo "running unit tests with postgres on port $POSTGRES_PORT"
+    echo "running unit tests with postgres"
+    CONFIG_FILE=config-devel.toml
     run_unit_tests
-else
-    echo "postgres on port $POSTGRES_PORT is not running, skipping unit tests with postgres"
```
Sergey1011010, line 40: Maybe we need this code so that `make before_commit` can run tests on postgres in case postgres is in the system

quarckster: I propose to explicitly specify if you have running postgres locally. I made this change because  `registry.redhat.io/rhel8/go-toolset:1.13` doesn't have `netstat`.

Makefile
```diff
@@ -53,6 +53,9 @@ json-check: ## Check all JSONs for basic syntax
 	@echo "Run JSON checker"
 	python3 utils/json_check.py
 
+openapi-check:
```
Sergey1011010, line 56: Do we still have openapi-check in `make before_commit`?

quarckster: @Sergey1011010 I removed openapi check from the integration tests. It runs docker which obviously cannot be used easily from another container.

Sergey1011010: I mean that `make before_commit` should run locally the same checkers as are in CI. Not a big deal, we can fix it later.

quarckster: I agree, but using containers limits us which and how we can run the checks.

Jenkinsfile
```diff
@@ -0,0 +1,55 @@
...
+
+node {
+    pipelineUtils.cancelPriorBuilds()
+
+    pipelineUtils.runIfMasterOrPullReq {
+        runStages()
+    }
+}
+
+def getBaseCommit() {
```
quarckster, line 11: This is temporary here. I will put it into insights-pipeline-lib.

check_openapi.sh
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo "Testing OpenAPI specifications file"
+# shellcheck disable=2181
+
+if docker run --rm -v "${PWD}":/local/:Z openapitools/openapi-generator-cli validate -i ./local/openapi.json; then
```
JoseLSegura, line 19: Have you seen that `openapi-generator-cli` tool is available in Pypi? Maybe it will be easy to automate using a "pip install" than running a container

quarckster: That's interesting, I will give it a try.

quarckster: @JoseLSegura 
> This project checks the maven repository once a day for a new version and will publish this new version automatically as a python package.

I'm afraid it still requires java :)

JoseLSegura: Dammit Java!!!

Makefile
quarckster: @Sergey1011010 I added it to `make before_commit`

PR #813, opened 2020-05-25
Some refactoring

Sergey1011010:
# Description

Just moved rules related endpoints to a different file and updated the DB structure in docs

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/docs/db_structure.md
+++ b/docs/db_structure.md
@@ -26,6 +26,7 @@ CREATE TABLE report (
     report          VARCHAR NOT NULL,
     reported_at     TIMESTAMP,
     last_checked_at TIMESTAMP,
+    kafka_offset BIGINT  NOT NULL DEFAULT 0,
     PRIMARY KEY(org_id, cluster)
 )
 ```
@@ -64,6 +65,7 @@ CREATE TABLE rule_error_key (
     publish_date  TIMESTAMP NOT NULL,
     active        BOOLEAN NOT NULL,
     generic       VARCHAR NOT NULL,
+    tags VARCHAR NOT NULL DEFAULT '',
     PRIMARY KEY(error_key, rule_module)
 )
 ```
--- a/server/rules.go
+++ b/server/rules.go
@@ -1,0 +1,316 @@
+package server
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+
+	"github.com/rs/zerolog/log"
+
+	"github.com/RedHatInsights/insights-operator-utils/responses"
+	"github.com/RedHatInsights/insights-results-aggregator/storage"
+	"github.com/RedHatInsights/insights-results-aggregator/types"
+)
+
+// likeRule likes the rule for current user
+func (server *HTTPServer) likeRule(writer http.ResponseWriter, request *http.Request) {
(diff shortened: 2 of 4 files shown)


Sergey1011010:
@tisnik it's part of 772, just wanted to make huge piece of refactoring a different PR to not distract from real changes

server/rules.go
```diff
@@ -0,0 +1,316 @@
...
+
+	"github.com/rs/zerolog/log"
+
+	"github.com/RedHatInsights/insights-operator-utils/responses"
+	"github.com/RedHatInsights/insights-results-aggregator/storage"
+	"github.com/RedHatInsights/insights-results-aggregator/types"
+)
+
+// likeRule likes the rule for current user
+func (server *HTTPServer) likeRule(writer http.ResponseWriter, request *http.Request) {
```
bond95, line 16: Maybe better to move all this vote stuff, to separate file?

docs/db_structure.md
```diff
@@ -64,6 +65,7 @@ CREATE TABLE rule_error_key (
     publish_date  TIMESTAMP NOT NULL,
     active        BOOLEAN NOT NULL,
     generic       VARCHAR NOT NULL,
+    tags VARCHAR NOT NULL DEFAULT '',
```
bond95, line 68: Do we need this? Because before it worked well and no new features that use this `tags` was not added.

docs/db_structure.md
```diff
@@ -26,6 +26,7 @@ CREATE TABLE report (
     report          VARCHAR NOT NULL,
     reported_at     TIMESTAMP,
     last_checked_at TIMESTAMP,
+    kafka_offset BIGINT  NOT NULL DEFAULT 0,
```
bond95, line 29: Same question here.

docs/db_structure.md
Sergey1011010: This did exist, I just updated the docs to be the same as the state of the DB after all applied migrations :)

docs/db_structure.md
Sergey1011010: Same answer here :)

docs/db_structure.md
bond95: Oh, ok, thank you =)

server/rules.go
Sergey1011010: separate file for ~40 lines?

tisnik:
> @tisnik it's part of 772, just wanted to make huge piece of refactoring a different PR to not distract from real changes

ok got it

server/rules.go
bond95: yes, I think logically it would be better. Like in file `rules` you have operations with rules, in file `votes` you have operations with votes.

bond95: And it will be ~80 lines.

Sergey1011010: then we would need file for disabling/enabling rules and creating/deleting which is too much I think.

bond95: well, about `disabling/enabling rules and file creating/deleting` I'm not sure, because for example number functions that are working with votes is 6, I think it's enough for separate file, no?

Sergey1011010: I don't think so, anyway we can move it later in different PR

PR #867, opened 2020-06-11
fix-773

### LINKED ISSUE #773: Log uses JSON mixed with colorized plain text messages when logger/debug is set to true
Settings:
```
[logging]
debug = true
log_level = ""
logging_to_cloud_watch_enabled = false
```

Log:

```
3:24PM INF Version: *not set* type=init
3:24PM INF Build time: *not set* type=init
3:24PM INF Branch: *not set* type=init
3:24PM INF Commit: *not set* type=init
3:24PM INF Making connection to data storage, driver=sqlite3WithHooks datasource=./aggregator.db
{"level":"debug","type":"SQL","message":"query `SELECT COUNT(*) FROM migration_info;` with params `[]`\n"}
3:24PM ERR unable to check DB migration version error="no such table: migration_info"
3:24PM INF Closing connection to data storage
3:24PM INF database preparation exited with error code 1
```

Sergey1011010:
# Description

Fixes #773

## Type of change

## Testing steps

`make before_commit` & test that SQL logs are displayed correctly


docs/pycco.css
```diff
@@ -1,3 +1,19 @@
+/**
```
tisnik, line 1: that is not our code at all :)

natiiix: Then we must somehow trick the license script into skipping this file.

tisnik: yeah

tisnik: @natiiix IMHO the script should accept any license, because we might have this problem with other files as well.

Sergey1011010: fixed and created an issue to not have this problem in the future https://github.com/RedHatInsights/insights-results-aggregator/issues/885

PR #870, opened 2020-06-17
Makefile improvement

### LINKED ISSUE #841: Improve Makefile
The `Makefile` right now is not taking advantage of rule dependencies, so some testing-related rules are building several times the same exact sources, increasing the execution time of that tests.

### Acceptance criteria
- Measure how many time takes to run all the tests and how many will take with the proposed improvement
- `Makefile` interface for users is kept (no Make rule name changes)

joselsegura:
# Description

Minor changes on Makefile to take advantage of file dependencies from Make

Fixes #841 

## Type of change

## Testing steps
Regular CI


joselsegura:
> Does it make sense to `:=` a constant variable? thinking
> Anyways, a cleaner `Makefile` is always welcome! +1

I was reading a bit the Make documentation and the only difference between using `=` and `:=` is when there are some "expansions" inside the value, so probably it will behave the same.

Saying that, the POSIX standard uses `::=` instead `:=`, which is something I didn't know :-D

PR #878, opened 2020-06-18
Update stable

quarckster:


joselsegura:
@Sergey1011010 is the `report` endpoint already working on smart-proxy and deployed? When it is, we should push a new stable version for aggregator!

quarckster:
@joselsegura smart-proxy hasn't been deployed yet. I will start the process.

PR #884, opened 2020-06-21
refactored-the-whole-universe

### LINKED ISSUE #883: Cannot request report endpoint when auth is disabled
Currently, the report endpoint returns the feedback and "disabled" toggle for each hit rule by a cluster, as can be seen in [`openapi.json`](https://github.com/RedHatInsights/insights-results-aggregator/blob/84d4f8dfe71bc39052d3eac1f1a12e47c8a28a6b/openapi.json#L223).

For that reason, an "user id" is needed to perform the request.

The problem is that currently the "user id" can only be retrieved when the Authentication is enabled, as it takes the user id from the authentication tokens and put it on the request's context.

It is needed that the "user id" can be taken from another source, for example, the request parameters

Sergey1011010:
## Type of change

Fixes #883 

Blocked by https://github.com/RedHatInsights/insights-results-aggregator-data/pull/5
// Blocked by https://github.com/RedHatInsights/insights-results-aggregator-utils/pull/93
Blocked by https://github.com/RedHatInsights/insights-operator-utils/pull/62

- Documentation update

## Testing steps

`make before_commit`


openapi.json
```diff
@@ -151,7 +151,7 @@
         ]
       }
     },
-    "/report/{orgId}/{clusterId}": {
+    "/organizations/{orgId}/clusters/{clusterId}/users/{userId}/report": {
```
quarckster, line 154: Why do we need userId and organization in the url? AFAIK we should get this data in `x-rh-identity` header.

joselsegura: I guess it is related to move the authentication to the smart proxy service. Aggregator won't have authentication anymore, so the smart proxy (or manual testing) can provide the user id in the URL instead of authentication

Sergey1011010: @quarckster because of the new architecture. We don't want to pass extra information such as authentication token to the aggregator, so we pass user_id directly in parameters.

If you want to see endpoint for users, it's here https://github.com/RedHatInsights/insights-results-smart-proxy/pull/39/files#diff-e2493ceba9636a2a265230ffacb30ed8

PR #906, opened 2020-06-29
Refactor for getAuthToken function

### LINKED ISSUE #856: The function is not simple "getter", so it have to has different name
The function sends response to client - should not be named `getSomething`:

https://github.com/RedHatInsights/insights-results-aggregator/blob/d4a58bd3edddea32b6594f3e233d3674a8c13fc7/server/auth.go#L137

joselsegura:
# Description

Refactores getauthToken function to be a full-getter. Now, the response is sent, if needed, from the caller

Fixes #856 

## Type of change

## Testing steps

Regular CI


joselsegura:
Just updated the godoc generated documentation

Issue #397, opened 2020-03-25
Implement a check of maximum length of user's feedback


Sergey1011010:
Now there's no any.

PR #909, opened 2020-07-01
Added feedback message validation

bond95:
# Description

Added message check in request body

Fixes https://github.com/RedHatInsights/insights-results-aggregator/issues/397

## Type of change

## Testing steps

Check through sending request

### DIFF
--- a/server/vote_endpoints.go
+++ b/server/vote_endpoints.go
@@ -15,13 +15,39 @@
 package server
 
 import (
-	"net/http"
-
+	"encoding/json"
 	"github.com/rs/zerolog/log"
+	"io"
+	"net/http"
 
 	"github.com/RedHatInsights/insights-operator-utils/responses"
 	"github.com/RedHatInsights/insights-results-aggregator/types"
 )
+
+// readFeedbackRequestBody parse request body and return object with message in it
+func (server *HTTPServer) readFeedbackRequestBody(writer http.ResponseWriter, request *http.Request) (types.FeedbackRequest, bool) {
+	var feedback types.FeedbackRequest
+
+	err := json.NewDecoder(request.Body).Decode(&feedback)
+	switch {
+	case err == io.EOF:
+		feedback.Message = ""
+		return feedback, true
+	case err != nil:
+		handleServerError(writer, err)
+		return feedback, false
+	}
+
+	if len(feedback.Message) > 250 {
+		handleServerError(writer, &types.ValidationError{
+			ErrString:  "String is longer then 250",
+			ParamName:  "message",
+			ParamValue: feedback.Message,
+		})
+		return feedback, false
+	}
+	return feedback, true
+}
 
 // likeRule likes the rule for current user
 func (server *HTTPServer) likeRule(writer http.ResponseWriter, request *http.Request) {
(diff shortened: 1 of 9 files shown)


server/vote_endpoints.go
```diff
@@ -15,14 +15,40 @@
...
+	"encoding/json"
 	"github.com/rs/zerolog/log"
+	"io"
+	"net/http"
 
 	"github.com/RedHatInsights/insights-operator-utils/responses"
 	"github.com/RedHatInsights/insights-results-aggregator/types"
 )
 
+// readFeedbackRequestBody likes the rule for current user
```
Sergey1011010, line 27: Description is not correct

server/vote_endpoints.go
```diff
@@ -15,14 +15,40 @@
...
+	switch {
+	case err == io.EOF:
+		feedback.Message = ""
+		return feedback, true
+	case err != nil:
+		handleServerError(writer, err)
+		return feedback, false
+	}
+
+	if len(feedback.Message) > 250 {
```
Sergey1011010, line 41: I would do it on storage side, but ok

server/vote_endpoints.go
```diff
@@ -15,14 +15,40 @@
...
+		handleServerError(writer, &types.ValidationError{
+			ErrString:  "String is longer then 250",
+			ParamName:  "message",
+			ParamValue: feedback.Message,
```
Sergey1011010, line 45: Maybe it would be better to take the first 250 characters of the message

server/vote_endpoints.go
joselsegura: I agree, it should be done in the storage

storage/storage_rules_test.go
```diff
@@ -470,10 +470,10 @@ func TestDBStorageGetVotes(t *testing.T) {
...
 	helpers.FailOnError(t, mockStorage.VoteOnRule(
-		testdata.ClusterName, testdata.Rule1ID, testdata.UserID, types.UserVoteLike,
+		testdata.ClusterName, testdata.Rule1ID, testdata.UserID, types.UserVoteLike, "",
 	))
 	helpers.FailOnError(t, mockStorage.VoteOnRule(
-		testdata.ClusterName, testdata.Rule2ID, testdata.UserID, types.UserVoteDislike,
+		testdata.ClusterName, testdata.Rule2ID, testdata.UserID, types.UserVoteDislike, "",
 	))
 
 	feedbacks, err := mockStorage.GetUserFeedbackOnRules(
```
joselsegura, line 479: Maybe you can add some test with a feedback string longer than 250 to check that the appropriate status code is returned

joselsegura:
Just out of curiosity, are we doing something with feedback right now? I only see the length check and nothing else.

server/vote_endpoints.go
bond95: I don't think it should be done on storage side, because checking validity of request data is usually doing before any operations with storage.

server/vote_endpoints.go
natiiix: If the front-end doesn't officially allow >250 characters to be submitted, then it may be okay to throw it away. Otherwise, we should probably get the first 250 characters and somehow indicate that the rest of it has been truncated.

bond95: Yep, agree, truncate is not very good idea. Much better just return error.

storage/storage_rules_test.go
bond95: Added test for long message

server/vote_endpoints.go
Sergey1011010: @bond95 what if the user send, say, 1MB of the data in message?

Sergey1011010:
@joselsegura you're right, we're not collecting it now, it was planned tho

PR #926, opened 2020-07-17
Fixed typo

### LINKED ISSUE #913: Typo: then -> than
https://github.com/RedHatInsights/insights-results-aggregator/blob/266f309058cd16c154387022df03231458bd8f49/server/vote_endpoints.go#L43

current error string is: "String is longer then 250"
better is to use: "String is longer than 250 bytes"

joselsegura:
# Description

Minor typo

Fixes #913 

## Type of change

## Testing steps

CI


tisnik:
@joselsegura please also add "bytes" as it is very important in this context.

joselsegura:
> @joselsegura please also add "bytes" as it is very important in this context.

Done

PR #944, opened 2020-07-21
increased timeout for integration tests

### LINKED ISSUE #879: Timeout for making coverage for integration tests should be bigger
2 minutes timeout is ok for other kinds of tests, but not for integration, cuz they are run manually by external program, so we need to change it here https://github.com/RedHatInsights/insights-results-aggregator/blob/master/make-coverage.sh#L58 to something like an hour.

Sergey1011010:
# Description

Fixes #879

## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/make-coverage.sh
+++ b/make-coverage.sh
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 TIMEOUT=2m
+TIMEOUT_INTEGRATION=1440m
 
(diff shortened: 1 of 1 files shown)


Sergey1011010:
@joselsegura 24 hours actually, but that's intentional

PR #949, opened 2020-07-23
Update stable

quarckster:


quarckster:
One test failed due to old version of archive generator. The PR can be safely merged.

PR #961, opened 2020-07-29
Splited report into two tables

bond95:
# Description

For now it's just consume report and save it into two tables.

## Type of change

## Testing steps

Run tests


migration/mig_0013_add_rule_table.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"database/sql"
+
+	"github.com/RedHatInsights/insights-results-aggregator/types"
+)
+
+var mig0013AddRuleTable = Migration{
+	StepUp: func(tx *sql.Tx, driver types.DBDriver) error {
+		_, err := tx.Exec(`
+		CREATE TABLE rule (
```
tisnik, line 30: Probably it might be good to think about better name of this table.

storage/storage.go
```diff
@@ -424,14 +440,57 @@ func (storage DBStorage) WriteReportForCluster(
...
+	kafkaOffset types.KafkaOffset,
+) error {
+	// Skip writing the report if it isn't newer than a report
+	// that is already in the database for the same cluster.
+	if oldLastChecked, exists := storage.clustersLastChecked[clusterName]; exists && !lastCheckedTime.After(oldLastChecked) {
+		return types.ErrOldReport
+	}
+
 	// Check if there is a more recent report for the cluster already in the database.
-	rows, err := tx.Query(
```
tisnik, line 428: don't removing this operation from transaction cause prohibition of atomic upsert operation?

quarckster:
@tisnik ack

migration/mig_0013_add_rule_table.go
bond95: Maybe `hit_rule`?

tisnik: or `rule_hit`, what you prefer

storage/storage.go
bond95: I didn't quite understand your question.

tisnik: I mean if the query is called outside transaction, there's no guarantee you can trust it's resultset later. Some other process can update database in meantime.

bond95: hmm... ok, it makes sense

consumer/processing.go
```diff
@@ -211,35 +212,41 @@ func checkReportStructure(r Report) error {
 }
 
 // parseMessage tries to parse incoming message and read all required attributes from it
-func parseMessage(messageValue []byte) (incomingMessage, error) {
+func parseMessage(messageValue []byte) (incomingMessage, []types.ReportItem, error) {
```
joselsegura, line 215: Why not including "rules" as a field of the `incomingMessage` type instead changing the number of returned valus, and in fact the rules array is part of the incoming message

quarckster:
@bond95 please rebase

PR #978, opened 2020-07-30
Change produced_messages metric description

### LINKED ISSUE #923: Update the description of produced_messages metric in prometheus.md
Update the description of `produced_messages` metric in `prometheus.md`

This metric contain counts of messages written to Payload tracker, but on Grafana it might confuse people (it can be read as number of messages written into `ccx.ocp.results` topic).

joselsegura:
# Description

In order to avoid misunderstandings of the meaning of the `produced_messages` metric, update docs, comments and current code.

Fixes #923 

## Type of change

## Testing steps

Regular CI

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -27,7 +27,7 @@ limitations under the License.
 //
 // last_checked_timestamp_lag_minutes - shows how slow we get messages from clusters
 //
-// produced_messages - total number of produced messages
+// produced_messages - total number of produced messages sent to Payload Tracker's Kafka topic
 //
 // written_reports - total number of reports written into the storage (cache)
 //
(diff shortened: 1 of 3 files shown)


metrics/metrics.go
```diff
@@ -27,7 +27,7 @@ limitations under the License.
 //
 // last_checked_timestamp_lag_minutes - shows how slow we get messages from clusters
 //
-// produced_messages - total number of produced messages
+// produced_messages - total number of produced messages sent to Payload Tracker's Kafka topic
```
Sergey1011010, line 30: I thought it's not about payload tracker..

joselsegura: As far as I can tell, it is. The only place where we are increasing the value of this counter is un `producer.produceMessage` method (OBVIOUS :-D), and the sent message is a `PayloadTrackerMessage`

tisnik: Yup, that's the reason why I was scared by this metric on Grafana :)

PR #982, opened 2020-07-31
Adding timeout to broker configuration

### LINKED ISSUE #924: Add Sarama library configuration
As `insights-operator-results` operator, I want to being able to configure some Kafka client-side parameters available in the library.

The current implementation of the aggregator's consumer allow to pass a `sarama.Config`, but it is not used, just passing a `nil` pointer and letting `sarama` to use its default configuration.

Some of the configuration options includes networking timeouts, that could be very useful to being able to configure from the aggregator's side.

### Acceptance criteria
* Identify relevant/useful configurations that Sarama allows
* Add this configurations to the `broker` configuration section and create the `sarama` structs according to those configuration parameters instead always using the default values

joselsegura:
# Description

Added a `timeout` configuration for Kafka client

Fixes #924 

## Type of change

## Testing steps

Regular CI


consumer/consumer.go
```diff
@@ -81,6 +81,12 @@ func NewWithSaramaConfig(
 	if saramaConfig == nil {
 		saramaConfig = sarama.NewConfig()
 		saramaConfig.Version = sarama.V0_10_2_0
+
+		if brokerCfg.Timeout > 0 {
```
Sergey1011010, line 85: It's applied only to the default config(if there was no saramaConfig passed. Is it expected behaviour?

joselsegura: I think so, because if you are passing a `sarama.Config` you can define the timeouts directly without using our configuration at all.

Anyway, we are always using the default configuration at the moment.

PR #1013, opened 2020-08-25
increased coverage to 85%

Sergey1011010:
## Type of change

## Testing steps

`make before_commit`

### DIFF
--- a/check_coverage.sh
+++ b/check_coverage.sh
@@ -20,6 +20,8 @@ RED_BG=$(tput setab 1)
 GREEN_BG=$(tput setab 2)
 NC=$(tput sgr0) # No Color
 
+make test || exit 1
+
 go_tool_cover_output=$(go tool cover -func=coverage.out)
 
 if [[ $* == *verbose* ]]; then
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/DATA-DOG/go-sqlmock"
+	operator_utils_types "github.com/RedHatInsights/insights-operator-utils/types"
 	"github.com/RedHatInsights/insights-results-aggregator-data/testdata"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -39,15 +40,6 @@ import (
 	"github.com/RedHatInsights/insights-results-aggregator/tests/helpers"
 	"github.com/RedHatInsights/insights-results-aggregator/types"
 )
-
-var config = server.Configuration{
-	Address:                      ":8080",
-	APIPrefix:                    "/api/test/",
-	APISpecFile:                  "openapi.json",
-	Debug:                        true,
-	Auth:                         false,
-	MaximumFeedbackMessageLength: 255,
-}
 
 func init() {
 	zerolog.SetGlobalLevel(zerolog.WarnLevel)
@@ -67,7 +59,7 @@ func checkResponseCode(t *testing.T, expected, actual int) {
 }
 
 func TestListOfClustersForNonExistingOrganization(t *testing.T) {
-	helpers.AssertAPIRequest(t, nil, &config, &helpers.APIRequest{
+	helpers.AssertAPIRequest(t, nil, nil, &helpers.APIRequest{
 		Method:       http.MethodGet,
 		Endpoint:     server.ClustersForOrganizationEndpoint,
 		EndpointArgs: []interface{}{1},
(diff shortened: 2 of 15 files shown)


tisnik (approved):
Looks good @Sergey1011010

server/server_test.go
```diff
@@ -722,3 +714,117 @@ func TestHTTPServer_deleteClusters_BadClusterName(t *testing.T) {
...
+	})
+
+	feedback, err := mockStorage.GetUserFeedbackOnRuleDisable(testdata.ClusterName, testdata.Rule1ID, testdata.UserID)
+	helpers.FailOnError(t, err)
+
+	assert.Equal(t, expectedFeedback, feedback.Message)
+}
+
+func TestHTTPServer_SaveDisableFeedback_Error(t *testing.T) {
+	t.Run("BadClusterName", func(t *testing.T) {
```
bond95, line 746: Usually for each test we write own function, so to be consistent I suggest split it.

server/server_test.go
```diff
@@ -722,3 +714,117 @@ func TestHTTPServer_deleteClusters_BadClusterName(t *testing.T) {
...
+		helpers.AssertAPIRequest(t, nil, nil, &helpers.APIRequest{
+			Method:       http.MethodPost,
+			Endpoint:     server.DisableRuleFeedbackEndpoint,
+			EndpointArgs: []interface{}{testdata.BadClusterName, testdata.Rule1ID, testdata.UserID},
+			Body:         fmt.Sprintf(`{"message": ""}`),
```
bond95, line 751: Why you use here `Sprintf`?

server/server_test.go
```diff
@@ -722,3 +714,117 @@ func TestHTTPServer_deleteClusters_BadClusterName(t *testing.T) {
...
+		err := mockStorage.WriteReportForCluster(
+			testdata.OrgID, testdata.ClusterName, testdata.Report3Rules, testdata.LastCheckedAt, testdata.KafkaOffset,
+		)
+		helpers.FailOnError(t, err)
+
+		helpers.AssertAPIRequest(t, mockStorage, &helpers.DefaultServerConfigAuth, &helpers.APIRequest{
+			Method:       http.MethodPost,
+			Endpoint:     server.DisableRuleFeedbackEndpoint,
+			EndpointArgs: []interface{}{testdata.ClusterName, testdata.Rule1ID, testdata.UserID},
+			Body:         fmt.Sprintf(`{"message": ""}`),
```
bond95, line 773: Same question here

server/server_test.go
```diff
@@ -722,3 +714,117 @@ func TestHTTPServer_deleteClusters_BadClusterName(t *testing.T) {
...
+		)
+		helpers.FailOnError(t, err)
+
+		closer()
+
+		helpers.AssertAPIRequest(t, mockStorage, nil, &helpers.APIRequest{
+			Method:       http.MethodPost,
+			Endpoint:     server.DisableRuleFeedbackEndpoint,
+			EndpointArgs: []interface{}{testdata.ClusterName, testdata.Rule1ID, testdata.UserID},
+			Body:         fmt.Sprint(`{"message": ""}`),
```
bond95, line 824: And here

bond95 (changes requested):

server/server_test.go
Sergey1011010: Why to copy paste the name each time when we can use a function as a namespace?

server/server_test.go
Sergey1011010: Thanks, didn't notice

check_coverage.sh
```diff
@@ -20,6 +20,8 @@ RED_BG=$(tput setab 1)
 GREEN_BG=$(tput setab 2)
 NC=$(tput sgr0) # No Color
 
+make test || exit 1
```
joselsegura, line 23: It is a nitpick, so you can ignore it if you want

This will make that `make before_commit` runs the tests twice (`before_commit` depends on `test` and will execute this script too).

I think it is better to just check if coverage.out exists instead of running "make test" without any condition

joselsegura (approved):
LGTM in general, just a nitpick

check_coverage.sh
Sergey1011010: The purpose is not to be sure that coverage.out exists, but to be sure it's up to date(before this workaround, the script did not actually check coverage as the name states)

joselsegura: It's ok for me.

The only thing is that I just realized that the Jenkins configuration first run "make test-postgres" and then perform the check_coverage, so the coverage is checked using a different test. With your change, Jenkins will run first "make test-postgres", and then check_coverage, that will run "make test", so the coverage.out file will be overwritten. Probably it is not a problem at all, just guessing

server/server_test.go
bond95: In consistency sake. @Sergey1011010 @tisnik do we want to change test style?

tisnik: @bond95 @Sergey1011010 what's the problem?

bond95: @tisnik inconsistency with other tests.

tisnik: +1 @bond95

Sergey1011010: @tisnik no problems, splitted to multiple functions

tisnik: TY @Sergey1011010

check_coverage.sh
Sergey1011010: Actually, I fixed because it seemed to be a problem :D

bond95 (approved):

PR #1022, opened 2020-09-07
Updated comments in `json_check.py`

tisnik:
# Description

Updated comments in `json_check.py` + added link in GitHub pages.

## Type of change

- Documentation update

## Testing steps

N/A


tisnik:
> LGTM
> 
> It would be useful to add json_check.py and other tooling to be autogenerated by the Makefile magic

Good idea. Will do it. Thank you!

PR #1037, opened 2020-10-08
fix-1032

### LINKED ISSUE #1032: DB writer unable to recover after Kafka outage
It's possible the situation when the db writer stops to consume messages from a Kafka topic. If Kafka had some outage db writer puts the following to the log:
```
ccx.ocp.results/0: dial tcp 10.130.20.14:9092: connect: connection refused
kafka: error while consuming ccx.ocp.results/0: EOF
Broker initialization error
```
Only the service restart can fix that. 

The db writer should be able to recover after Kafka outages.

Sergey1011010:
Fixes #1032 

## Type of change

- Refactored some spaghetti 

## Testing steps

`make before_commit` & test all the cases with a real kafka

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -168,17 +168,23 @@ func startService() int {
 
 	errorGroup := new(errgroup.Group)
 
-	errorGroup.Go(func() error {
-		defer cancel()
+	brokerConf := conf.GetBrokerConfiguration()
+	// if broker is disabled, simply don't start it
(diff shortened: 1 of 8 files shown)


tisnik (approved):
Ah it looks good, thank you!

joselsegura (changes requested):
LGTM, but please update documentation with `make godoc`

joselsegura (approved):

PR #1042, opened 2020-10-13
First try to update OpenAPI JSON file

tisnik:
# Description

First try to update OpenAPI JSON file

## Type of change

- Documentation update

## Testing steps

N/A

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -274,7 +274,7 @@
         "parameters": [
           {
             "name": "clusterIDs",
-            "in": "body",
+            "in": "path",
             "required": true,
             "description": "List of cluster IDs. Each ID must conform to UUID format. An example: `34c3ecc5-624a-49a5-bab8-4fdc5e51a266.",
             "schema": {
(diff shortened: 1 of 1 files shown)


openapi.json
```diff
@@ -266,6 +266,131 @@
...
+            "name": "clusterIDs",
+            "in": "body",
+            "required": true,
+            "description": "List of cluster IDs. Each ID must conform to UUID format. An example: `34c3ecc5-624a-49a5-bab8-4fdc5e51a266.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "minLength": 36,
+                "maxLength": 36,
```
quarckster, line 285: @tisnik  Did you keep in mind that URL length is around 2000 chars? Therefore a requester should implement a sort of pagination if it needs a big list of clusters.

openapi.json
```diff
@@ -266,6 +266,131 @@
         ]
       }
     },
+    "/clusters/reports/": {
```
quarckster, line 269: Should it be here `orgId` as well?

openapi.json
tisnik: we don't want to use URL for providing any parameters (for this particular call). Everything is sent via request body.

openapi.json
tisnik: unfortunatelly ACM has no way (right now) how to provide orgId :/

openapi.json
quarckster: What request body? Your endpoint is for GET that means all parameters go to URL.

openapi.json
quarckster: This looks like a big flaw in the authorization. This endpoint can be used to access data that are not authorized for the requester.

openapi.json
tisnik: semantically it is GET, and the parameters are send in body. IIRC it is valid according to rfc7231 (the old http 1.1 is no longer valid) and that approach is used for example in ElasticSearch. We can change it to POST, of course, but then it will have wrong semantic.

(yes, old HTTP is broken in many way)

Sergey1011010 (approved):

quarckster (approved):

Bee-lee (approved):
:+1:

joselsegura (approved):
lgtm

PR #1043, opened 2020-10-13
fixed-unauthorized-error

Sergey1011010:
Blocked by https://github.com/RedHatInsights/insights-operator-utils/pull/117

## Type of change

## Testing steps

`make before_commit`


quarckster (approved):

tisnik (approved):
More precise, TY

Bee-lee:
I think that 401 and 403 are actually kind of unfortunately named and this is not the correct usage. I think that Unauthorized should only be used for "missing auth token", but I guess we;re not building a perfect API anyway

https://tools.ietf.org/html/rfc7235#section-3.1
https://tools.ietf.org/html/rfc7231#section-6.5.3

PR #1053, opened 2020-10-20
New endpoint to retrieve multiple results

tisnik:
# Description

New endpoint to retrieve multiple results

## Type of change

- Documentation update

## Testing steps

N/A

### DIFF
--- a/docs/packages/server/endpoints.html
+++ b/docs/packages/server/endpoints.html
@@ -176,7 +176,7 @@ limitations under the License.</p>
       </tr>
       
       <tr class="section">
-	<td class="doc"><p>ReportEndpoint returns report for provided {organization}</p>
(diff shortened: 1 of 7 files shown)


quarckster:
@tisnik do I understand right that is an implementation of #1042?

tisnik:
> @tisnik do I understand right that is an implementation of #1042?

yes

quarckster (approved):

Sergey1011010 (approved):
We could probably reuse some functions from router-utils like https://github.com/RedHatInsights/insights-operator-utils/blob/4c1dd92d7b3ea8a80058c4d14469c7830f7105d9/http/router_utils.go#L221, but it doesn't really matter. Looks great, thanks :)

PR #1095, opened 2021-01-28
Incoming message version validation

### LINKED ISSUE #1091: Check version field in incoming messages from `ccx.ocp.results`
In order to be able to use proper validation for incoming messages in `ccx.ocp.results` we need to check version field that can be paired with schema validator. We don't have to use major.minor.patch versions (i.e. semantic versioning), just major number might be enough.

The new field should be named `version` and it should contain an integer constant `1`.

Possible location where the verification can be added:
https://github.com/RedHatInsights/insights-results-aggregator/blob/661a7c997a04021fbc7f9d14e80043a0f5864e8d/consumer/processing.go#L115

See also https://github.com/RedHatInsights/ccx-data-pipeline/issues/385

epapbak:
# Description

Incoming messages from `ccx.ocp.results`shall now include the Version field.
If received message's version does not match the expected version, a warning
is logged. 

Fixes #1091 

## Type of change

- Unit tests
- Documentation update

## Testing steps

Added new UTs in consumer/consumer_test.go
Run existing UTs


quarckster (approved):

consumer/consts.go
```diff
@@ -29,4 +33,8 @@ const (
 	clusterKey = "cluster"
 	// key for duration message type used in structured log messages
 	durationKey = "duration"
+	// key for data schema version message type used in structured log messages
+	versionKey = "version"
+	// currently supported data schema version
```
joselsegura, line 38: Probably linters will freak out with this comment :-P

tisnik (approved):
LGTM

consumer/consts.go
epapbak: They are, they are =)

joselsegura:
LGTM

consumer/consts.go
tisnik: :rescue_worker_helmet:

joselsegura (approved):

Bee-lee (approved):
LGTM

PR #1150, opened 2021-04-09
Make disabling rules cluster-wide / user-independent

### LINKED ISSUE #1126: Store information about rule enable/disable just once for given cluster
Currently every user can enable or disable rule for given cluster separately.
This behaviour needs to be changed - it's needed to store this info just once.

### Table `cluster_rule_toggle`

```
   Column    |            Type             | Modifiers
-------------+-----------------------------+-----------
 cluster_id  | character varying           | not null
 rule_id     | character varying           | not null
 user_id     | character varying           | not null
 disabled    | smallint                    | not null
 disabled_at | timestamp without time zone |
 enabled_at  | timestamp without time zone |
 updated_at  | timestamp without time zone | not null
Indexes:
    "cluster_rule_toggle_pkey" PRIMARY KEY, btree (cluster_id, rule_id, user_id)
Check constraints:
    "cluster_rule_toggle_disabled_check" CHECK (disabled >= 0 AND disabled <= 1)
```

My suggestion is NOT to change the database, but change the logic for inserting/updating the rule in DB.
Dunno what to do with PRIMARY KEY - `user_id` might be removed.

### Table `cluster_user_rule_disable_feedback`

```
   Column   |            Type             | Modifiers
------------+-----------------------------+-----------
 cluster_id | character varying           | not null
 user_id    | character varying           | not null
 rule_id    | character varying           | not null
 message    | character varying           | not null
 added_at   | timestamp without time zone | not null
 updated_at | timestamp without time zone | not null
Indexes:
    "cluster_user_rule_disable_feedback_pkey" PRIMARY KEY, btree (cluster_id, user_id, rule_id)
```

This is IMHO perfectly ok, as every user can comment separately.

Bee-lee:
# Description
This PR makes the ability to disable rules/recommendations/health checks cluster-wide and user-independent.
I tried doing it without DB migration, but there was a problem with the PK, so I added a new migration. It doesn't drop the `user_id` column, only makes it nullable and alters the PK.

There is a [test case](https://github.com/Bee-lee/insights-results-aggregator/blob/0d431b9eecca327bf34ad7d7c7ff0cdeb75bcaf9/server/server_read_report_test.go#L249) that "mimics" some of the [behavioral tests](https://github.com/RedHatInsights/insights-behavioral-spec/blob/main/features/OCM_disable_rule.feature), without the UI part of course, so please check that one out

PS: this really shouldn't have taken me so long, I forgot how a lot of the stuff works internally with smart-proxy etc ... took a lot of notes so that it doesn't happen again.

Fixes  #1126

## Type of change

- Documentation update

## Testing steps
tested alongside new version of smart-proxy, `make test`/`before_commit`

## Checklist


migration/mig_0014_modify_cluster_rule_toggle.go
```diff
@@ -0,0 +1,99 @@
...
+	`,
+)
+
+var mig0014ModifyClusterRuleToggle = Migration{
+	StepUp: func(tx *sql.Tx, driver types.DBDriver) error {
+		if driver == types.DBDriverPostgres {
+			return mig0014ModifyClusterRuleToggleAlter.StepUp(tx, driver)
+		}
+
+		return mig0014ModifyClusterRuleToggleGeneral.StepUp(tx, driver)
```
joselsegura, line 89: I don't fully understand the migrations process, so probably my question is stupid (feel free to ignore it).

If I understand this correctly, when running in Postgresl, only the "ALTER TABLE"s will be executed, but IIRC the table cluster_rule_toggle doesn't exist in the previous migration of the DB, and only if the driver is != of Postgres (SQLite), the table will be created.

Am I wrong?

server/router_utils.go
```diff
@@ -196,3 +196,30 @@ func (server *HTTPServer) readClusterRuleUserParams(
 
 	return clusterID, ruleID, userID, true
 }
+
+// readClusterRuleParams gets cluster_name and rule_id from current request
+func (server *HTTPServer) readClusterRuleParams(
+	writer http.ResponseWriter, request *http.Request,
+) (types.ClusterName, types.RuleID, bool) {
```
joselsegura, line 203: Nitpick: As you are returning 3 values here, probably using named return values will improve the readability.

joselsegura (approved):
LGTM, just a doubt about the migration and some nitpick.

Great work!!!

tisnik (approved):
LGTM

migration/mig_0014_modify_cluster_rule_toggle.go
Bee-lee: What do you mean @joselsegura? https://github.com/RedHatInsights/insights-results-aggregator/blob/master/migration/mig_0007_create_cluster_rule_toggle.go

The non-psql way, the mig0014ModifyClusterRuleToggleGeneral is only there because SQLite is too dumb to understand the ALTER TABLE, so there is this helper function by Sergey (NewUpdateTableMigration) that creates a new table with the new schema and copies the data from the old table.

joselsegura: Ouch, Ok. I missed that the cluster_rule_toggle already exist on our database, my fault

server/router_utils.go
Bee-lee: definitely a good idea to do that, we have a lot of such functions in aggregator, even worse than this one. Might be a good candidate for some maintenance/readability tasks.

PR #1153, opened 2021-04-13
[CCXDEV-3976] add haberdasher configuration

dpensi:
# Description

this PR enables the insights-result-aggregator and the insights-result-db-writer to run with haberdasher logging

## Type of change

## Testing steps
build the docker image and than run it, verify that your first log line is like so:
`2021/04/13 13:54:07 Initializing haberdasher.`

and no subsequent errors are shown


tisnik (approved):
It looks ok, thank you!

joselsegura (approved):
Probably the `conf` package update is not necessary to make haberdasher work, but it's ok to upgrade the dependency :-)

PR #1162, opened 2021-05-06
do not panic on misconfiguration

dpensi:
# Description

Do not panic if the configuration for logging is wrong

## Type of change

## Checklist


tisnik (approved):
Let's try

joselsegura (approved):
Let's try, but I assume that if logging fails to initialize, probably we won't be able to see anything about this log

Bee-lee (approved):

PR #1179, opened 2021-06-17
fix upsert query which wasn't ready for a cluster transferring to different org

Bee-lee:
# Description
When a cluster got transferred to a different org (e.g. via OCM), our pipeline was basically not ready for it, only expecting a `org_id` + `cluster_id` ON CONFLICT. However when we received a cluster report that had the same cluster_id but a different org_id, the ON CONFLICT was not triggered, the db-writer tried to perform a simple INSERT and failed because of the enforced `org_id` + `cluster_id` uniqueness.

This PR changes the ON CONFLICT behaviour so that when the service receives a report with a new org_id, it's able to handle it.

No other changes to SELECT queries are IMO necessary because the constraint is handled on DB side, so a unique cluster ID always = 1 org. **This means we're now relying on cluster IDs (UUIDs) being unique, which I think is probably safe to do**. And if we experience a cluster_id duplicate, I'll be damned :) 

Fixes https://issues.redhat.com/browse/CCXDEV-5142

## Type of change

## Testing steps
tried the behaviour locally by modifying a report's org_id in the database and trying to upload the archive again (which now wouldn't trigger the ON CONFLICT) and it works as expected = the cluster gets transferred to the new org, which is now able to retrieve the report and the old org is not able to access to report

## Checklist

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,3 +1,17 @@
+# Copyright 2021 Red Hat, Inc
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


joselsegura (approved):

tisnik (approved):
Very nice!

PR #1184, opened 2021-07-13
[CCXDEV-3462] Add error_key to vote/feedback related tables

joselsegura:
# Description

The database only stores the rule_id when a user enable, disable, gives feedback or vote a rule hit result. This results on that a rule with several possible error keys can only be stored as a whole, not possible to enable/disable an specific hit on that rule.

The commit adds support on the database to store the error_key and enables the related endpoints to store and retrieve data from the updated tables.

Fixes insights-results-smart-proxy#340 

## Type of change

## Testing steps

Regular CI

## Checklist


server/router_utils.go
```diff
@@ -197,29 +197,34 @@ func (server *HTTPServer) readClusterRuleUserParams(
...
-// readClusterRuleParams gets cluster_name and rule_id from current request
+// readClusterRuleParams gets cluster_name, rule_id and error_key from current request
 func (server *HTTPServer) readClusterRuleParams(
 	writer http.ResponseWriter, request *http.Request,
-) (types.ClusterName, types.RuleID, bool) {
+) (types.ClusterName, types.RuleID, types.ErrorKey, bool) {
 	clusterID, successful := readClusterName(writer, request)
 	if !successful {
-		return "", "", false
+		return "", "", "", false
```
tisnik, line 206: the right time to use named return values, WDYT?

server/router_utils.go
```diff
@@ -197,29 +197,34 @@ func (server *HTTPServer) readClusterRuleUserParams(
...
 	}
 
 	ruleID, successful := readRuleID(writer, request)
 	if !successful {
-		return "", "", false
+		return "", "", "", false
```
tisnik, line 211: dtto

server/router_utils.go
```diff
@@ -197,29 +197,34 @@ func (server *HTTPServer) readClusterRuleUserParams(
...
 	clusterExists, err := server.Storage.DoesClusterExist(clusterID)
 	if err != nil {
 		handleServerError(writer, err)
-		return "", "", false
+		return "", "", "", false
 	}
 	if !clusterExists {
 		handleServerError(writer, &types.ItemNotFoundError{ItemID: clusterID})
-		return "", "", false
+		return "", "", "", false
```
tisnik, line 226: dtto

tisnik (approved):
Very nice one! I just found the function with four return values quite unreadable ATM, that's all.

server/router_utils.go
joselsegura: you are right, new commit incoming

tisnik (approved):
LGTM!!!

Bee-lee (approved):
Looks good to me! Very nice.

We just have to make sure it isn't displayed weird in OCM UI, a user can run into a situation where he disables a rule and then the cluster hits a different error key of the same rule and it will pop back again since we're now selecting by `error_key` too.
Since I believe we're showing a general message in the lists, and also we're planning to accommodate for "general" summary/resolution/reason as well as ones specific to the error keys, it could look like a bug by having the rule duplicated or seemingly re-enabled before going to the details of them.

joselsegura:
> Looks good to me! Very nice.
> 
> We just have to make sure it isn't displayed weird in OCM UI, a user can run into a situation where he disables a rule and then the cluster hits a different error key of the same rule and it will pop back again since we're now selecting by `error_key` too.
> Since I believe we're showing a general message in the lists, and also we're planning to accommodate for "general" summary/resolution/reason as well as ones specific to the error keys, it could look like a bug by having the rule duplicated or seemingly re-enabled before going to the details of them.

Yes, we need to check it very carefully. I suppose that different error keys will be shown differently in OCM UI (and TBH most of the rules that we are running in the pipeline only have 1 error key)

joselsegura:
I'm closing this PR at the moment because I just realised that I need to update one more table, which its database schema was not touched from the migration part

joselsegura:
Reopened the PR. Updates on missing tables and proper manually testing about migration in Postgres DB

Bee-lee (approved):
LGTM still

dpensi (approved):

epapbak (approved):
lgtm

PR #1196, opened 2021-08-02
Add new create_table operation to migrations for recommendations_view…

epapbak:
Recommendations view PoC

# Description

Draft PR with the changes for new “Recommendations view” functionality .

## Type of change

- REST API tests
- Benchmarks (no changes in the code)
- Documentation update

## Testing steps

TBD

## Checklist


storage/storage.go
```diff
@@ -729,6 +753,70 @@ func (storage DBStorage) WriteReportForCluster(
...
+
+	var deletedRows int
+	var insertedRows int
+
+	err = func(tx *sql.Tx) error {
+
+		// Delete current recommendations for the cluster
+		result, err := tx.Exec(
+			"DELETE FROM recommendations WHERE cluster_id = $1;", clusterName)
+		err = types.ConvertDBError(err, []interface{}{clusterName})
```
epapbak, line 779: Not sure this is needed with the DELETE operation, but used it just in case

migration/mig_0016_add_recommendations_table.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+var mig0016AddRecommendationsTable = Migration{
+	StepUp: func(tx *sql.Tx, driver types.DBDriver) error {
+		_, err := tx.Exec(`
+		CREATE TABLE recommendations (
+			cluster_id      VARCHAR NOT NULL,
+			rule_fqdn       VARCHAR NOT NULL,
+			error_key		VARCHAR NOT NULL,
+			PRIMARY KEY(cluster_id)
```
tisnik, line 29: does it mean that there will be just one record per cluster?

storage/storage.go
```diff
@@ -729,6 +765,63 @@ func (storage DBStorage) WriteReportForCluster(
...
+			return err
+		}
+
+		// As the documentation says:
+		// RowsAffected returns the number of rows affected by an
+		// update, insert, or delete. Not every database or database
+		// driver may support this.
+		// So we might run in a scenario where we don't have metrics
+		// if the driver doesn't help.
+		deleted, err := result.RowsAffected()
```
tisnik, line 800: that's ok, PQ can do it (we have unit tests that works with it correctly)

migration/mig_0016_add_recommendations_table.go
```diff
@@ -0,0 +1,106 @@
...
+				break
+			} else if err != nil {
+				return err
+			}
+
+			err = writeRulesFromReportToRecommendations(tx, clusterID, report)
+			if err != nil {
+				return err
+			}
+		}
```
tisnik, line 69: that's interesting solution. Just would like to know if this solution based on cursor is better than classic iteration over result set.

storage/storage.go
tisnik: hmm you want to do `UPSERT` right? Probably better to do INSERT with ON CONFLICT clause?

tisnik (commented):
Pretty good progress!!!

storage/storage.go
epapbak: I didn't see it that way. I thought for each report we would "reset" the corresponding clusters, so upsert could leave behind some unwanted entries. But maybe I didn't understand how the recommendations would work

migration/mig_0016_add_recommendations_table.go
epapbak: True, true. I don't even remember why I changed it to just cluster_id :)

migration/mig_0016_add_recommendations_table.go
epapbak: I went for cursor solution because I'm not sure how many rows can be in the report table. If it is not a big number, classic iteration should be the best solution

storage/storage.go
tisnik: Ah you are correct @epapbak

migration/mig_0016_add_recommendations_table.go
tisnik: :)

migration/mig_0016_add_recommendations_table.go
tisnik: not a problem, really. Btw number of clusters is pretty stable - around 3500, so let's count with 10000.

Bee-lee:
looks like a good piece of work so far! +1 for the new metrics, cool idea!

PR #1220, opened 2021-09-02
add disabled_at and disable_feedback

Bee-lee:
# Description
I'm having problems with my local ingress so I want to see this in QA. 
The format is the same as our other timestamps, so the frontend should be expecting it -- also wanna see this on UI.

CCXDEV-5290

## Type of change

- 
## Testing steps
`make before_commit`

## Checklist

### DIFF
--- a/server/rules.go
+++ b/server/rules.go
@@ -160,6 +160,7 @@ func (server HTTPServer) getFeedbackAndTogglesOnRule(
 		rule.Disabled = false
 	} else {
 		rule.Disabled = ruleToggle.Disabled == storage.RuleToggleDisable
+		rule.DisabledAt = types.Timestamp(ruleToggle.DisabledAt.Time.UTC().Format(time.RFC3339))
 	}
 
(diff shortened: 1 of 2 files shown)


tisnik (approved):
That one is short one :)

server/rules.go
```diff
@@ -160,6 +160,7 @@ func (server HTTPServer) getFeedbackAndTogglesOnRule(
 		rule.Disabled = false
 	} else {
```
dpensi, line 161: wrong indentation?

server/rules.go
```diff
@@ -168,6 +169,7 @@ func (server HTTPServer) getFeedbackAndTogglesOnRule(
 		rule.UserVote = types.UserVoteNone
```
dpensi, line 169: same here, wrong indentation?

dpensi (approved):

server/rules.go
Bee-lee: it looks ok in my ide, probably just line breaks on github?

joselsegura (approved):

PR #1229, opened 2021-09-08
Drop recommendation table and recreate it with org_id column

epapbak:
# Description

This commit is necessary to update the database in stage environment.
The "DROP TABLE" operation will be removed from the migration code afterwards.

## Type of change

## Testing steps

- Tested in local DB
- `make test-postgres` passes
- `make integration_tests` passes

## Checklist

### DIFF
--- a/migration/mig_0016_add_recommendations_table.go
+++ b/migration/mig_0016_add_recommendations_table.go
@@ -37,9 +37,16 @@ var mig0016AddRecommendationsTable = Migration{
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


tisnik (approved):
ok, but looks pretty dangerous :)

Bee-lee (approved):
ok

epapbak:
> ok, but looks pretty dangerous :)

it is, it is.
That's why the plan is to push this to stage, and push another commit removing the "drop" line as fast as possible =)

PR #1240, opened 2021-09-14
Add org ID to recommendation table's composite key

epapbak:
# Description

In production environment, the current primary key cannot be created as there might be some old duplicated cluster IDs. 

## Type of change

## Testing steps

Local test

## Checklist


dpensi (approved):

tisnik (approved):
Seems ok (not sure about downgrading though)

epapbak:
> Seems ok (not sure about downgrading though)

AFAIK, `DROP TABLE` cannot be used to drop a table that is referenced by a FOREIGN KEY constraint. That's not the case here. But I'll take a look as postgres docs this afternoon, see if there is something to be wary off in the downgrade step.

PR #1281, opened 2021-09-27
System wide rule disable storage methods + unit tests

tisnik:
# Description

System wide rule disable storage methods + unit tests

## Type of change

## Testing steps

Done on CI

## Checklist

### DIFF
--- a/storage/rule_disable.go
+++ b/storage/rule_disable.go
@@ -1,0 +1,227 @@
+// Copyright 2021 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


dpensi (approved):

storage/rule_disable.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+// DisableRuleSystemWide disables the selected rule for all clusters visible to
+// given user
+func (storage DBStorage) DisableRuleSystemWide(
+	orgID types.OrgID, userID types.UserID,
+	ruleID types.RuleID, errorKey types.ErrorKey,
+	justification string) error {
+
+	now := time.Now()
```
Bee-lee, line 33: is this saved/returned in the expected format like other timestamps? I don't remember the exact reason, but the [sql.NullTime](https://github.com/RedHatInsights/insights-results-aggregator/blob/36781f0e7c6b4ba2fa51fb2e17fc13b08e93f4e6/storage/rule_toggle.go#L61) was used for the toggling, but I guess it was only done so because the columns could've been null, which is not possible here

tisnik: yup it is ok there

Bee-lee (approved):
LGTM

PR #1285, opened 2021-09-29
Modify recommendation table

### LINKED ISSUE #1274: Recommendations insertion does not behave properly when report with 0 hits is received
When a report with 0 rule hits is received, the SQL query used for inserting recommendations in the `recommendation` table is not correctly formed but still sent to the DB. What should be done in that case is check if the DB has entries for the corresponding cluster and remove them

epapbak:
# Description
- Fix the `recommendation` table's `rule_fqdn` column through migration19, and add UT in `migration/actual_migrations_test.go`
    - REGEXP_REPLACE(rule_fqdn, '\.(?!.*\|)(?!.*\.)', '|') replaces the last `.` found in the rule FQDN if the character `|` is not found in the current value 
- Fix some comments
- Fix insertion of empty recommendations (which fails due to bad SQL statement) when no rule hit in processed report
- Add `rule_id` column whose content is `rule.module|error_key`.
- Add `created_at` column in `recommendation` table, with default UTC timestamp in existing records, and update `storage.insertRecommendations` function to insert the timestamp.   

Fixes #1274 

## Type of change

- Bug fix
- New feature
- Unit tests

## Testing steps

new UTs + local testing

## Checklist

### DIFF
--- a/migration/actual_migrations_test.go
+++ b/migration/actual_migrations_test.go
@@ -374,7 +374,7 @@ func TestMigration16(t *testing.T) {
 	helpers.FailOnError(t, err)
 
 	_, err = db.Exec(`
-		INSERT INTO recommendations (org_id, cluster, rule_fqdn, error_key)
+		INSERT INTO recommendation (org_id, cluster_id, rule_fqdn, error_key)
 		VALUES ($1, $2, $3, $4)
 	`,
 		testdata.OrgID,
@@ -382,13 +382,14 @@ func TestMigration16(t *testing.T) {
 		testdata.Rule1Name,
 		testdata.ErrorKey1,
 	)
-	assert.EqualError(t, err, "no such table: recommendations")
+	assert.Error(t, err, `Expected error since recommendation table does not exist yet`)
+	assert.Contains(t, err.Error(), `relation "recommendation" does not exist`)
 
 	err = migration.SetDBVersion(db, dbDriver, 16)
 	helpers.FailOnError(t, err)
 
 	_, err = db.Exec(`
-		INSERT INTO recommendations (org_id, cluster, rule_fqdn, error_key)
+		INSERT INTO recommendation (org_id, cluster_id, rule_fqdn, error_key)
 		VALUES ($1, $2, $3, $4)
 	`,
 		testdata.OrgID,
@@ -397,7 +398,6 @@ func TestMigration16(t *testing.T) {
 		testdata.ErrorKey1,
 	)
 	helpers.FailOnError(t, err)
-
 }
 
 func TestMigration19(t *testing.T) {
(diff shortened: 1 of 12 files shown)


migration/mig_0016_add_recommendations_table.go
```diff
@@ -43,8 +43,8 @@ var mig0016AddRecommendationsTable = Migration{
 				AS SELECT
 					org_id,
 					cluster_id,
-					REGEXP_REPLACE(rule_fqdn, 'report$', error_key) AS rule_fqdn,
-					error_key
+					REGEXP_REPLACE(rule_fqdn, '.report$', CONCAT('|', error_key)) AS rule_fqdn,
```
tisnik, line 46: IIRC the `error_key` should not be appended, just store `rule_fqdn` as is (IMHO)

migration/mig_0019_modify_recommendation_table.go
```diff
@@ -0,0 +1,38 @@
...
+var mig0019ModifyRecommendationTable = Migration{
+	StepUp: func(tx *sql.Tx, driver types.DBDriver) error {
+		if driver != types.DBDriverPostgres {
+			return nil
+		}
+
+		// Fix rule_fqdn value for records created in migration 16
+		_, err := tx.Exec(`
+			UPDATE recommendation
+				SET rule_fqdn = REGEXP_REPLACE(rule_fqdn, '\.(?!.*\|)(?!.*\.)', '|');
```
tisnik, line 30: what does this replace do might be commented, because TBH it is not much readable ;)

epapbak: TBH, it is not readable at all 😄

migration/mig_0016_add_recommendations_table.go
epapbak: I'm removing this change from the PR, as fixing this removes the sense of mig_0019

Bee-lee:
I sent you a DM, but I'll just repeat here -- I just realized that it would likely be worth it to introduce a 3rd column combining them in the `|` format...

Hear me out :D The problem is that we don't really have any content structs or anything optimized for this notation, but we have to use it, so IMHO it would be best to introduce a new map of rule content in smart-proxy which will use the `|` as key so we can access the content directly without iterating over rules/error keys. That way we should be optimized everywhere while also being "future proof" regarding the possibility of selecting only by FQDN or error key.

What do you think?

migration/mig_0016_add_recommendations_table.go
```diff
@@ -31,19 +31,18 @@ var mig0016AddRecommendationsTable = Migration{
...
 		}
 
+		// Create recommendation table using records from rule_hit table
 		_, err := tx.Exec(`
 			CREATE TABLE recommendation
 				AS SELECT
 					org_id,
 					cluster_id,
-					REGEXP_REPLACE(rule_fqdn, 'report$', error_key) AS rule_fqdn,
+					REGEXP_REPLACE(rule_fqdn, '.report$', '') AS rule_fqdn,
```
joselsegura, line 45: Is it a good idea to modify older migrations that are, probably, already applied to the DB structure? It won't break anything in local deployments, but maybe it can fail when deploying to prod

joselsegura (changes requested):
Just asking for a clarification, the rest of the code looks ok

migration/mig_0016_add_recommendations_table.go
epapbak: It won't do anything =)
But I thought I removed this change from the PR. Thanks for noticing!

migration/actual_migrations_test.go
```diff
@@ -370,35 +371,132 @@ func TestMigration16(t *testing.T) {
...
+		`,
+		testdata.OrgID,
+		testdata.ClusterName,
+		incorrectRuleFQDN,
+		testdata.ErrorKey1,
+	)
+	helpers.FailOnError(t, err)
 
+	_, err = db.Exec(`
+		INSERT INTO recommendation (org_id, cluster_id, rule_fqdn, error_key)
```
tisnik, line 439: not `rule_id`?

migration/actual_migrations_test.go
```diff
@@ -370,35 +371,132 @@ func TestMigration16(t *testing.T) {
...
+	helpers.FailOnError(t, err)
+
+	var (
+		ruleFQDN string
+		ruleID   string
+	)
+
+	err = db.QueryRow(`
+			SELECT
+				rule_fqdn, rule_id
```
tisnik, line 459: `error_key`?

Bee-lee (approved):
looks good to me. :cool:

storage/export_test.go
```diff
@@ -56,16 +56,16 @@ func SetClustersLastChecked(storage *DBStorage, cluster types.ClusterName, lastC
 	storage.clustersLastChecked[cluster] = lastChecked
 }
 
-func InsertRecommendations(storage *DBStorage, orgID types.OrgID, clusterName types.ClusterName, report types.ReportRules) error {
+func InsertRecommendations(storage *DBStorage, orgID types.OrgID, clusterName types.ClusterName, report types.ReportRules) (int, error) {
```
tisnik, line 59: `inserted int`

tisnik (commented):
LGTM in overall, just few comments

tisnik:
@epapbak also please "squash and merge"

epapbak:
> @epapbak also please "squash and merge"

Definitely =)

migration/mig_0019_modify_recommendation_table.go
epapbak: Commented

migration/actual_migrations_test.go
epapbak: Not sure what's the question =)


I'm not testing `error_key` in migration 19 because at that point it already exists in the table.

joselsegura (approved):

PR #1289, opened 2021-10-01
add aggregator part for /rule/ recommendations list

Bee-lee:
# Description
Aggregator part for GET /rule/ which returns a list of recommendations for the active clusters in the organization (given the AMS API is alive) with a number the number of clusters its impacting.

ready for changes in the recommendation in progress by @epapbak

TODO: bump utils repo, modify unit tests

## Type of change

- Documentation update

## Checklist


Bee-lee:
@tisnik
1. this is aggregator, there is only one API version :) 
2. as far as I know the number of params there can be up to short int, do you think we can expect some org to have over 32K clusters??

PR #1296, opened 2021-10-13
Clusters details per rule

epapbak:
# Description

New endpoint for getting a list of clusters affected by a given recommendation.

The endpoint expects a query in the format `/rules/{rule_selector}/organizations/{org_id}/users/{user_id}/clusters_detail` (see openapi.json file for details about the parameters) and returns a JSON list of objects containing a cluster UUID, similar to the following:
```
{
    "data": [
        {"cluster":"5d5892d3-1f74-4ccf-91af-548dfc9767aa"},
        {"cluster":"5d5892d3-1f74-4ccf-91af-548dfc976ccc"}
    ],
    "meta": {
        "count":2,
        "component":"ccx_rules_plugin.name",
        "error_key":"ccx_rules_error_key"
    },
    "status":"ok"
}```

Additional data is expected to be added within the JSON object.

## Type of change

## Testing steps

Added UTs (and adapted existing UTs after insights-operator-utils version update)
Manual testing of new endpoint + CI

## Checklist

### DIFF
--- a/storage/rule_feedback.go
+++ b/storage/rule_feedback.go
@@ -233,7 +233,7 @@ func (storage DBStorage) GetUserFeedbackOnRules(
 		FROM cluster_rule_user_feedback
 		WHERE cluster_id = $1 AND rule_id in (%v) AND user_id = $2`
 
-	whereInStatement := "'" + strings.Join([]string(ruleIDs), "','") + "'"
+	whereInStatement := inClauseFromSlice(ruleIDs)
 	query = fmt.Sprintf(query, whereInStatement)
 
 	rows, err := storage.connection.Query(query, clusterID, userID)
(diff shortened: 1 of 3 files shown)


epapbak:
> is it supposed to support disabled rules?

If I understood correctly, "not right now". Which means this implementation is valid, but will need to be adapted if at some point there's the need.

Bee-lee:
Also keep in mind that the list has to be cross checked with the list of active clusters retrieved from the AMS API in smart-proxy.

Right now, we'll be selecting clusters based on our DB, which is including stale and inactive clusters that we mustn't return to the requester. Maybe it would be better to retrieve the list from AMS first and pass that to aggregator to select from? WDYT?

PR #1317, opened 2021-11-02
Fix clusters_detail per recommendation SQL query

epapbak:
# Description

Comparing activeClusters slice with nil was not a good idea, and the resulting SQL query for the `clusters_detail` endpoint was incorrect as we were entering the wrong condition with no data, resulting in the following query:  `SELECT cluster_id FROM recommendation WHERE org_id = $1 AND rule_id = $2 AND cluster_id IN ('') ORDER BY cluster_id;`

## Type of change

## Testing steps

- Tested locally with smart-proxy calls to check that the endpoint now works as expected
- `make test` and `make test-postgres` pass
- CI

## Checklist


tisnik (approved):
Much better now!

juandspy (approved):

dpensi (approved):

PR #1327, opened 2021-11-09
 temporarily add EXPLAIN ANALYZE to recommendation list query 

Bee-lee:
# Description
let's see how it behaves and see if a manually created index would even be used by the engine

## Testing steps
ran locally and checked the EXPLAIN ANALYZE output

## Checklist


tisnik (approved):
ok

dpensi (approved):

PR #1335, opened 2021-11-30
Adding aggregator part for OCP Advisor Clusters view (Milestone 2)

Bee-lee:
# Description
Needed for https://issues.redhat.com/browse/CCXDEV-6486
1st part of Clusters View in OCP Adivsor, the new endpoint returns a map of cluster IDs and a list of hitting rule IDs for each cluster.

Fixes CCXDEV-6486

## Type of change

- REST API tests

## Testing steps
`make test` currently doesn't pass against sqlite

## Checklist

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1287,6 +1287,99 @@
         }
       }
     },
+    "/clusters/organizations/{org_id}/users/{user_id}/recommendations": {
+      "post": {
+        "summary": "getClustersRecommendationsList retrieves all hitting recommendations for all clusters given in the POST body",
+        "operationId": "getClustersRecommendationsPost",
+        "description": "Recommendations will be retrieved based on the list of cluster IDs that is part of request body.",
+        "parameters": [
+          {
+            "name": "org_id",
+            "in": "path",
+            "description": "Organization ID represented as positive integer",
+            "required": true,
+            "schema": {
+              "type": "integer",
+              "format": "int64",
+              "minimum": 0
+            }
+          },
+          {
+            "name": "user_id",
+            "in": "path",
+            "required": true,
+            "description": "Numeric ID of the user. An example: `42`",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "description": "List of cluster IDs. An example: `34c3ecc5-624a-49a5-bab8-4fdc5e51a266.",
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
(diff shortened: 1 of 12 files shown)


tisnik (approved):
nice solution

server/reports.go
```diff
@@ -304,3 +304,45 @@ func (server *HTTPServer) getRecommendations(writer http.ResponseWriter, request
 		log.Error().Err(err).Msg(responseDataError)
 	}
 }
+
+// getClustersRecommendationsList retrieves all recommendations hitting for all clusters in the org
```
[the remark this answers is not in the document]
Bee-lee: yes, much better, thank you

openapi.json
```diff
@@ -1287,6 +1287,99 @@
...
+                  "type": "object",
+                  "properties": {
+                    "clusters": {
+                      "type": "object",
+                      "additionalProperties": {
+                        "type": "object",
+                        "properties": {
+                          "created_at": {
+                            "type": "string",
+                            "description": "The cluster ID in UUID format."
```
epapbak, line 1342: created_at -> "The cluster ID in UUID format."

One of them needs fixing

Bee-lee: good catch, TYVM

openapi.json
```diff
@@ -1287,6 +1287,99 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "clusters": {
+                      "type": "object",
+                      "additionalProperties": {
+                        "type": "object",
+                        "properties": {
```
epapbak, line 1339: Did you discuss with UI if we want the cluster_name here too? (as we are doing in the cluster_details endpoint)?

epapbak (approved):
LGTM!

openapi.json
```diff
@@ -1287,6 +1287,99 @@
...
+                          "created_at": {
+                            "type": "string",
+                            "description": "The cluster ID in UUID format."
+                          },
+                          "recommendations": {
+                            "type": "array",
+                            "items": {
+                              "type": "string",
+                              "description": "The recommendation ID.",
```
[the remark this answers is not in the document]
Bee-lee: you're right, thx. We're now even since now you caught me copy pasting :)

openapi.json
Bee-lee: (discussed briefly on Slack)
this is not a proxied request, smart-proxy needs to match this data from aggregator with some data from rule content.
and considering we need to iterate over the list of clusters from AMS in smart-proxy because we need to return even clusters without hits, I think it's not needed to pass the cluster names to aggregator

PR #1338, opened 2021-12-01
fix sqlite/psql different behaviour

Bee-lee:
# Description
the `created_at` column is behaving differently when being selected by SQLite driver and PSQL driver. 
PostgreSQL is able to autoparse the column into time.Time directly whereas SQLite needs manual time parse.

This PR fixes the behaviour for both drivers, it's not the prettiest solution in the world, but we shouldn't even test on SQLite when all environments use postgres.

Fixes # (issue)

## Type of change

## Testing steps
`make before_commit` + calling the endpoint from smart-proxy with both SQLite and PostgreSQL running

## Checklist

### DIFF
--- a/storage/consts.go
+++ b/storage/consts.go
@@ -30,5 +30,5 @@ const (
 	// inClauseError when constructing IN clause fails
 	inClauseError = "error constructing WHERE IN clause"
 	// recommendationTimestampFormat represents the datetime format of the created_at of recommendation table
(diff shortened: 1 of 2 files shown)


tisnik (approved):
OK :timer_clock: :watch: :hourglass_flowing_sand:

juandspy (approved):

joselsegura (approved):

dpensi (approved):

PR #1382, opened 2021-12-22
Add dead letter queue

### LINKED ISSUE #1381: Create a dead letter queue
The goal is to create a Kafka topic to collect all messages that have not been correctly processed by aggregator.

JiriPapousek:
# Description

Created a dead letter queue to collect unprocessed messaged to dedicated Kafka topic.

Fixes #1381

## Type of change

## Testing steps

## Checklist

### DIFF
--- a/consumer/consumer.go
+++ b/consumer/consumer.go
@@ -59,7 +59,8 @@ type KafkaConsumer struct {
 	numberOfErrorsConsumingMessages      uint64
 	ready                                chan bool
 	cancel                               context.CancelFunc
-	payloadTrackerProducer               *producer.KafkaProducer
+	payloadTrackerProducer               *producer.PayloadTrackerProducer
+	deadLetterProducer                   *producer.DeadLetterProducer
 }
 
 // DefaultSaramaConfig is a config which will be used by default
@@ -94,9 +95,15 @@ func NewWithSaramaConfig(
 		return nil, err
 	}
 
-	payloadTrackerProducer, err := producer.New(brokerCfg)
+	payloadTrackerProducer, err := producer.NewPayloadTrackerProducer(brokerCfg)
 	if err != nil {
-		log.Error().Err(err).Msg("unable to construct producer")
+		log.Error().Err(err).Msg("unable to construct payload tracker producer")
+		return nil, err
+	}
+
+	deadLetterProducer, err := producer.NewDeadLetterProducer(brokerCfg)
+	if err != nil {
+		log.Error().Err(err).Msg("unable to construct dead letter producer")
 		return nil, err
 	}
 
@@ -108,6 +115,7 @@ func NewWithSaramaConfig(
 		numberOfErrorsConsumingMessages:      0,
 		ready:                                make(chan bool),
 		payloadTrackerProducer:               payloadTrackerProducer,
+		deadLetterProducer:                   deadLetterProducer,
 	}
 
 	return consumer, nil
@@ -210,6 +218,12 @@ func (consumer *KafkaConsumer) Close() error {
 	if consumer.payloadTrackerProducer != nil {
 		if err := consumer.payloadTrackerProducer.Close(); err != nil {
 			log.Error().Err(err).Msg("unable to close payload tracker Kafka producer")
+		}
+	}
+
+	if consumer.deadLetterProducer != nil {
+		if err := consumer.deadLetterProducer.Close(); err != nil {
+			log.Error().Err(err).Msg("unable to close dead letter Kafka producer")
 		}
 	}
 
--- a/producer/dead_letter_producer.go
+++ b/producer/dead_letter_producer.go
@@ -1,0 +1,72 @@
+/*
+Copyright © 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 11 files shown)


consumer/consumer.go
```diff
@@ -94,9 +95,15 @@ func NewWithSaramaConfig(
 		return nil, err
 	}
 
-	payloadTrackerProducer, err := producer.New(brokerCfg)
+	payloadTrackerProducer, err := producer.CreatePayloadTrackerProducer(brokerCfg)
```
juandspy, line 98: Maybe you can call it `NewPayloadTrackerProducer` instead of `CreatePayloadTrackerProducer`?

consumer/consumer.go
```diff
@@ -94,9 +95,15 @@ func NewWithSaramaConfig(
...
 	if err != nil {
-		log.Error().Err(err).Msg("unable to construct producer")
+		log.Error().Err(err).Msg("unable to construct payload tracker producer")
+		return nil, err
+	}
+
+	deadLetterProducer, err := producer.CreateDeadLetterProducer(brokerCfg)
```
juandspy, line 104: Also here. I think it is more common to use `New` instead of `Create` for instantiating new instances for a given type. Maybe I'm wrong

producer/dead_letter_producer.go
```diff
@@ -0,0 +1,70 @@
...
+// send) messages to properly configured Kafka broker.
+package producer
+
+import (
+	"github.com/RedHatInsights/insights-results-aggregator/broker"
+	"github.com/Shopify/sarama"
+	"github.com/rs/zerolog/log"
+)
+
+// CreateDeadLetterProducer is a variable for monkey patching in the unit tests
```
juandspy, line 27: I've just seen this. Why do you need a `Create` function?

juandspy (approved):
It looks great. I'd love to see some BDD tests for this as it doesn't look too difficult to test. It's probably out of scope and too much effort for this task.

producer/dead_letter_producer.go
JiriPapousek: The thing is, I need to create mock of DeadLetterProducer  and PayloadTrackerProducer for unit testing. But the variables containing those structs are instantiated in [KafkaConsumer constructor](https://github.com/RedHatInsights/insights-results-aggregator/blob/671708f49047e75bbe0a6a9c7173eecf92ebf983/consumer/consumer.go#L97), so it cannot be easily done e.g. by creating interface.

JiriPapousek: `CreateDeadLetterProducer` is in fact `NewDeadLetterProducer`, but is exchanged for custom function in unit tests.

producer/dead_letter_producer.go
```diff
@@ -0,0 +1,70 @@
...
+var CreateDeadLetterProducer = NewDeadLetterProducer
+
+// DeadLetterProducer is a producer for dead letter queue
+type DeadLetterProducer struct {
+	KafkaProducer KafkaProducer
+	Configuration broker.Configuration
+}
+
+// NewDeadLetterProducer constructs producer for payload tracker topic
+func NewDeadLetterProducer(brokerCfg broker.Configuration) (*DeadLetterProducer, error) {
```
dpensi, line 37: do we need to return an error if we return always nil?

producer/dead_letter_producer.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil
+	}
+
+	_, _, err := producer.KafkaProducer.produceMessage(msg.Value, producer.Configuration.DeadLetterQueueTopic)
+	if err != nil {
+		log.Error().Err(err).Msg("unable to produce message to dead letter queue")
+		return err
+	}
+
+	log.Info().Msg("message has been produced to dead letter queue")
```
dpensi, line 58: can we log the first two return values of ` producer.KafkaProducer.produceMessage` instead of ignoring them? Maybe just part of them, just I suggestion I can see how this may not be practical

producer/payload_tracker_producer.go
```diff
@@ -0,0 +1,105 @@
...
+	Configuration broker.Configuration
+}
+
+// NewPayloadTrackerProducer constructs producer for payload tracker topic
+func NewPayloadTrackerProducer(brokerCfg broker.Configuration) (*PayloadTrackerProducer, error) {
+	p, _ := New(brokerCfg)
+	return &PayloadTrackerProducer{
+		KafkaProducer: *p,
+		Configuration: brokerCfg,
+	}, nil
```
dpensi, line 46: same as above, an error is never actually returned... maybe we should not ignore the second return value on `New` in this case?

dpensi (approved):

producer/payload_tracker_producer.go
JiriPapousek: Yes, that is probably the case, I'll fix that :)

producer/dead_letter_producer.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright © 2020 Red Hat, Inc.
```
epapbak, line 2: 2022 =)

tisnik (approved):
LGTM, just pls. look at other comment, not need to repeat them :)

producer/dead_letter_producer.go
JiriPapousek: In the end I decided to rename the variable and assign the function directly to the variable without naming it.

PR #1385, opened 2022-01-05
[CCXDEV-5662]  consumer_test.go::TestKafkaConsumer_New_FindCoordinatorRequestError

juandspy:
# Description

This test is commented because it is failing. It checks that `consumer.New` returns an error if the `FindCoordinatorRequest` fails. However, I've checked Sarama source code and the only scenario when sarama.NewConsumerGroup returns an error is when [the client is closed](https://github.com/Shopify/sarama/blob/65f0fec86aabe011db77ad641d31fddf14f3ca41/consumer.go#L107). This test makes no sense as changing the `FindCoordinatorRequest` doesn't affect the client itself, so it won't return an error in this case.

Fixes #CCXDEV-5662

## Type of change

## Checklist

### DIFF
--- a/consumer/consumer_test.go
+++ b/consumer/consumer_test.go
@@ -398,29 +398,6 @@ func TestKafkaConsumer_New(t *testing.T) {
 	}, testCaseTimeLimit)
 }
 
-// TODO: fix with new groups consumer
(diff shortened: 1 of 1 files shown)


tisnik (approved):
ok

dpensi (approved):

joselsegura (approved):

PR #1387, opened 2022-01-18
Return lastSeen field in ListOfClustersForOrgSpecificRule's response

epapbak:
# Description

Include `last_seen` timestamp in cluster_details endpoint.

Fixes CCXDEV-6552

## Type of change

- Bump-up dependent library

## Testing steps

Updated UTs

## Checklist


epapbak:
This PR needs to include latest changes in https://github.com/RedHatInsights/insights-results-types

tisnik (commented):
is it really a string? And what format is expected in REST API -> UI?

epapbak:
> 

I went with string type for consistency with the timestamp field we already provide in /report endpoint (see https://github.com/RedHatInsights/insights-results-types/blob/master/types.go#L67 and https://github.com/RedHatInsights/insights-results-types/blob/master/types.go#L89, for example). But let me discuss with UI guys and ensure it would match what they expect.

epapbak:
> is it really a string? And what format is expected in REST API -> UI?

From UI perspective, they expect all dates/timestamps from backend to be a **string holding the timestamp in rfc3339**.

dpensi (approved):

tisnik (approved):
OK

PR #1435, opened 2022-02-22
add migration to clean-up cluster rule toggle and feedback tables

Bee-lee:
# Description
Because of recent changes, we have inconsistencies in `cluster_rule_toggle` and `cluster_user_rule_disable_feedback`. Some rows have `.report` suffix in Rule ID column and some don't.

This migration deletes all rows, where `rule_id` column doesn't end with `.report` suffix. 

**Important**:
- I checked Advisor UI on STAGE and the `.report` suffix is correctly included in all endpoints which require it.
  - this will need to be confirmed on PROD too before we deploy this migration, otherwise it makes no sense to run it yet.

Fixes CCXDEV-7321

## Type of change

## Testing steps
I'm unable to run storage-related unit tests locally, so let's see if I can write tests on the first try :D 

## Checklist

### DIFF
--- a/migration/actual_migrations_test.go
+++ b/migration/actual_migrations_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/RedHatInsights/insights-results-aggregator/migration"
+	"github.com/RedHatInsights/insights-results-aggregator/storage"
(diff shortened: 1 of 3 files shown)


tisnik (approved):
Still seems legit :p

PR #1446, opened 2022-03-11
Report endpoint metadata to include gathered_at

joselsegura:
# Description

Modify aggregator's report endpoint to retrieve and return the gathered time information, if any

## Type of change

## Testing steps
make before_commot

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -155,7 +155,10 @@ func (server *HTTPServer) readReportForCluster(writer http.ResponseWriter, reque
 		return
 	}
 
-	reports, lastChecked, _, err := server.Storage.ReadReportForCluster(orgID, clusterName)
+	reports, lastChecked, _, gatheredAt, err := server.Storage.ReadReportForCluster(
+		orgID,
+		clusterName,
+	)
 	if err != nil {
 		log.Error().Err(err).Msg("Unable to read report for cluster")
 		handleServerError(writer, err)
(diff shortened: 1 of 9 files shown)


dpensi (approved):

server/server.go
```diff
@@ -200,7 +204,10 @@ func (server *HTTPServer) readReportMetainfoForCluster(writer http.ResponseWrite
 		return
 	}
 
-	reports, lastChecked, storedAt, err := server.Storage.ReadReportForCluster(orgID, clusterName)
+	reports, lastChecked, storedAt, _, err := server.Storage.ReadReportForCluster(
```
juandspy, line 207: Why not using `gatheredAt` here?

joselsegura: I'm not implementing yet the update for this endpoint, it will be in a separated PR because it is common for both V1 and V2 endpoints in Smart Proxy. I need to split the logic in Smart Proxy first for that endpoint and then, I will be able to return the gathered_at field here too :-)

joselsegura: I'm not implementing yet the update for this endpoint, it will be in a separated PR because it is common for both V1 and V2 endpoints in Smart Proxy. I need to split the logic in Smart Proxy first for that endpoint and then, I will be able to return the gathered_at field here too :-)

tisnik (approved):
ok

PR #1566, opened 2022-03-23
Try different theme

tisnik:
# Description

Try different theme


joselsegura (approved):
Basically it reverts my last change IIRC. I though that the problem was that the theme repository was moved, but apparently it is not.

OTOH, if it is a totally new different repository, it might work. Let's see!

dpensi (approved):

PR #1587, opened 2022-04-11
Last checked at from report instead of recommendation table

Bee-lee:
# Description
In the Clusters list views in Advisor, we need to be able to show a `last_checked_at` timestamp. This is not possible using only `recommendation` table, because a cluster archive with 0 rule hits results in 0 rows in `recommendation` table, which results in a wrong behaviour, because a missing timestamp for a cluster means we hadn't received any archive yet.

This PR changes the query so it takes the timestamp from `report` table, which exists even if there are 0 rule hits and joins the recommendation table on `org_id` and `cluster_id`.

Fixes 7918

## Type of change

## Testing steps
tested all use cases (no archive, empty archive, >1 rule hit)

## Checklist

### DIFF
--- a/unit-tests.sh
+++ b/unit-tests.sh
@@ -17,7 +17,7 @@ STORAGE=$1
 
 function run_unit_tests() {
     # shellcheck disable=SC2046
-    if ! go test -timeout 3m -coverprofile coverage.out $(go list ./... | grep -v tests | tr '\n' ' ')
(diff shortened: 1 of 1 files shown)


dpensi (approved):

tisnik (approved):
LGTM

juandspy (approved):

PR #1606, opened 2022-05-04
Store report info

joselsegura:
# Description

Get the info reports from consumed reports and store the relevant "version info" into the database.
Added a "read" method to retrieve the version info from DB too

## Type of change

## Testing steps

By hand, some "before_commit" rules are broken locally due to Go version mismatch

## Checklist


dpensi (approved):

storage/queries.go
```diff
@@ -0,0 +1,65 @@
...
+		INSERT INTO report(org_id, cluster, report, reported_at, last_checked_at, kafka_offset, gathered_at)
+		VALUES ($1, $2, $3, $4, $5, $6, $7)
+		ON CONFLICT (cluster)
+		DO UPDATE SET org_id = $1, report = $3, reported_at = $4, last_checked_at = $5, kafka_offset = $6, gathered_at = $7
+	`
+}
+
+// getRuleHitInsertStatement method prepares DB statement to be used to write
+// rule FQDN + rule error key into rule_hit table for given cluster_id
+func (storage DBStorage) getRuleHitInsertStatement() string {
```
juandspy, line 37: what about this @joselsegura ?

storage/storage.go
```diff
@@ -760,22 +770,6 @@ func (storage DBStorage) GetLatestKafkaOffset() (types.KafkaOffset, error) {
...
-		INSERT INTO report(org_id, cluster, report, reported_at, last_checked_at, kafka_offset, gathered_at)
-		VALUES ($1, $2, $3, $4, $5, $6, $7)
-		ON CONFLICT (cluster)
-		DO UPDATE SET org_id = $1, report = $3, reported_at = $4, last_checked_at = $5, kafka_offset = $6, gathered_at = $7
-	`
-}
-
 // GetRuleHitInsertStatement method prepares DB statement to be used to write
 // rule FQDN + rule error key into rule_hit table for given cluster_id
 func (storage DBStorage) GetRuleHitInsertStatement(rules []types.ReportItem) string {
```
juandspy, line 775: Is this related to `func (storage DBStorage) getRuleHitInsertStatement() string` ? They have the same name.

juandspy (changes requested):
LGTM

storage/queries.go
joselsegura: At the same time I moved it to this file, the function was removed from `storage.go`, so the merge was not able to detect it. Thank you!

storage/storage.go
joselsegura: TBH I don't know, it is something that I didn't touch in this PR. The function body looks quite different.

PR #1611, opened 2022-05-09
Set an empty version if cannot be found in DB (second option)

juandspy:
# Description

In case there is no cluster version stored (which is very likely), return an empty string.

## Type of change

## Testing steps

Unit tests.

## Checklist


joselsegura (approved):

tisnik (approved):
LGTM

PR #1613, opened 2022-05-12
Using latest change on common types

joselsegura:
# Description

- Use the latest version of the HiitingClustersData with new "meta" field.
- Adapted previous implementation to the new approach
- Updated openapi spec

## Type of change

## Testing steps

Locally tested

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -350,26 +350,18 @@ func (server *HTTPServer) deleteClusters(writer http.ResponseWriter, request *ht
 	}
 }
 
-func (server *HTTPServer) addVersionToClusters(orgID types.OrgID, clusters []ctypes.HittingClustersData) ([]types.HittingClustersDataWithVersion, error) {
-	clustersWithVersion := []types.HittingClustersDataWithVersion{}
-
-	for _, cluster := range clusters {
-		version, err := server.Storage.ReadReportInfoForCluster(orgID, cluster.Cluster)
+func (server *HTTPServer) addVersionToClusters(orgID types.OrgID, clusters []ctypes.HittingClustersData) error {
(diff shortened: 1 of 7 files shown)


server/server.go
```diff
@@ -350,26 +350,18 @@ func (server *HTTPServer) deleteClusters(writer http.ResponseWriter, request *ht
...
 		}
 
-		clustersWithVersion = append(
-			clustersWithVersion,
-			types.HittingClustersDataWithVersion{
-				HittingClustersData: cluster,
-				Version:             version,
-			},
-		)
+		clusters[index].Meta = ctypes.ClusterMetadata{Version: version}
```
juandspy, line 361: If you want to modify the `clusters` you'll need to pass a pointer

joselsegura: AFAIK arrays/slices are passed by reference. If we access the element through its original reference, using the index, it will be modified in the original array.

When iterating over the elements, it gets copies instead, so the modifications are not in the original object

joselsegura: @tisnik pleasel could you verify if I'm right?

tisnik: yes, slice is basically struct with three members:
- pointer to an array
- actual length
- capacity

tisnik (approved):
LGTM, but pls make sure it won't break Smart Proxy

joselsegura:
> LGTM, but pls make sure it won't break Smart Proxy

I think that it cannot break anything: as it is using the previos version of the common types package, it will ignore the new "meta" attribute.

I will double check in stage, just in case

juandspy (approved):
nice simplification :)

PR #1614, opened 2022-05-16
Efectively store report info into db

joselsegura:
# Description

Report info is not being stored into DB. This commit fixed that

## Type of change

## Testing steps

Locally tested

## Checklist

### DIFF
--- a/consumer/processing.go
+++ b/consumer/processing.go
@@ -142,9 +142,9 @@ func checkMessageOrgInAllowList(consumer *KafkaConsumer, message *incomingMessag
 	return true, ""
 }
 
-func writeRecommendations(
-	consumer *KafkaConsumer, msg *sarama.ConsumerMessage, message incomingMessage, reportAsBytes []byte) (
-	time.Time, error) {
+func (consumer *KafkaConsumer) writeRecommendations(
+	msg *sarama.ConsumerMessage, message incomingMessage, reportAsBytes []byte,
+) (time.Time, error) {
 	err := consumer.Storage.WriteRecommendationsForCluster(
 		*message.Organization,
 		*message.ClusterName,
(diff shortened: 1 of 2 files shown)


tisnik (approved):
LGTM

juandspy (approved):

Bee-lee (approved):
ok

consumer/processing.go
```diff
@@ -161,6 +161,28 @@ func writeRecommendations(
...
+func (consumer *KafkaConsumer) writeInfoReport(
+	msg *sarama.ConsumerMessage, message incomingMessage, infoStoredAtTime time.Time,
+) error {
+	err := consumer.Storage.WriteReportInfoForCluster(
+		*message.Organization,
+		*message.ClusterName,
+		message.ParsedInfo,
+		infoStoredAtTime,
+	)
+	if err != nil {
```
dpensi, line 173: this is redundant https://go.dev/play/p/d2sRXgyH2ds you can compare nil and an error, maybe there's a way to leave gocyclo unchanged if we solve this?

dpensi (approved):
approving, left a couple of (unimportant) comments

consumer/processing.go
```diff
@@ -161,6 +161,28 @@ func writeRecommendations(
...
+		if err == types.ErrOldReport {
+			logMessageInfo(consumer, msg, message, "Skipping because a more recent info report already exists for this cluster")
+			return nil
+		}
+
+		logMessageError(consumer, msg, message, "Error writing info report to database", err)
+	}
+	logMessageInfo(consumer, msg, message, "Stored info report")
```
dpensi, line 181: in case `err != nil`

we will log both 
`"Error writing info report to database"`
And
`"Stored info report"`

you forgot a return?

consumer/processing.go
joselsegura: Maybe it will look cleaner, but it will still need to increase the gocyclo because the function with 10 gocyclo points is the `ParseMessage` one, and not this one

consumer/processing.go
```diff
@@ -161,6 +161,28 @@ func writeRecommendations(
...
+	err := consumer.Storage.WriteReportInfoForCluster(
+		*message.Organization,
+		*message.ClusterName,
+		message.ParsedInfo,
+		infoStoredAtTime,
+	)
+	if err != nil {
+		if err == types.ErrOldReport {
+			logMessageInfo(consumer, msg, message, "Skipping because a more recent info report already exists for this cluster")
+			return nil
```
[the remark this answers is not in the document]
joselsegura: No, when we skip a message we don't treat it as an error, just as something that should be logged. The same is done for the main report (probably it is not possible to reach this part of the code, but just in case)

dpensi (approved):

PR #953, opened 2020-07-23
Adding namespace to metrics and update documentation

joselsegura:
# Description

- Adding metrics namespace configuration parameter
- Use `insights-operator-utils` metrics with namespace too
- Update metrics related documentation and comments

Fixes #921 

## Type of change
- New feature (non-breaking change which adds functionality)
- Documentation update

## Testing steps

Regular CI

PR #1625, opened 2022-06-16
 persistent recommendations created date

dpensi:
# Description
makes created_at column of recommendation table persistent (doesn't get updated)

Fixes CCXDEV-7858

## Type of change

## Testing steps
make before_commit

## Checklist

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -940,6 +940,43 @@ func (storage DBStorage) insertRecommendations(
 
 }
 
+// getRecommendationsTime return the created_at value of recommendations
+// with given orgID and clusterName. If the value is not available or in
+// case of failure it returns an error
+func (storage DBStorage) getRecommendationsCreatedAt(
+	orgID types.OrgID,
+	clusterName types.ClusterName,
+) (
+	*types.Timestamp,
+	error) {
+
+	creationTimeRows, err := storage.connection.Query(
+		"SELECT created_at FROM recommendation WHERE org_id = $1 AND cluster_id = $2 LIMIT 1;", orgID, clusterName)
+	if err != nil {
+		log.Error().Err(err).Msg("error retrieving recommendation timestamp")
+		return nil, err
+	}
+
+	if exists := creationTimeRows.Next(); exists {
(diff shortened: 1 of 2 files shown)


tisnik (approved):
ok, but probably it would be better to put the code into its own function/method? WDYT?

storage/storage_test.go
```diff
@@ -1149,17 +1149,22 @@ func TestDBStorageWriteRecommendationForClusterAlreadyStoredAndDeleted(t *testin
...
+	expects.ExpectQuery(`SELECT created_at FROM recommendation`).
+		WillReturnRows(expects.NewRows([]string{"created_at"}).AddRow(time.Time{})).
+		RowsWillBeClosed()
 	expects.ExpectExec("DELETE FROM recommendation").
 		WillReturnResult(driver.RowsAffected(3))
 	expects.ExpectExec("INSERT INTO recommendation").
 		WillReturnResult(driver.ResultNoRows)
 	expects.ExpectCommit()
+	expects.ExpectClose()
+	expects.ExpectClose()
```
juandspy, line 1161: why is it necessary to expectClose twice?

juandspy (approved):

storage/storage.go
```diff
@@ -940,6 +940,44 @@ func (storage DBStorage) insertRecommendations(
...
+		return nil, err
+	}
+
+	if exists := creationTimeRows.Next(); exists {
+		var oldTime time.Time
+		if err := creationTimeRows.Scan(&oldTime); err != nil {
+			log.Error().Err(err).Msg("error scanning old time from recommendation")
+			return nil, err
+		}
+		newTime = types.Timestamp(oldTime.UTC().Format(time.RFC3339))
```
tisnik, line 967: `newTime := ` allows you to delete line #953

tisnik (approved):
LGTM

storage/storage_test.go
dpensi: it will fail if i remove one, and it will fail even if I don't call Close()

PR #1632, opened 2022-06-24
insert rule hit with created_at

dpensi:
# Description

ionsert rule hit entry with created_at column

Fixes CCXDEV-7857

## Type of change

## Testing steps
n/a

## Checklist

### DIFF
--- a/storage/rule_hit_preparation_test.go
+++ b/storage/rule_hit_preparation_test.go
@@ -15,12 +15,14 @@
 package storage_test
 
 import (
-	"github.com/stretchr/testify/assert"
 	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


tisnik (commented):
you don't have to delete+insert if rule hits are the same, probably just update the table instead?

dpensi:
> you don't have to delete+insert if rule hits are the same, probably just update the table instead?

delete + insert was an pre-existing behaviour, I think it's because we don't know how to update multiple rows 
what rule_fqdn, error_key, template_data and created_at we pair with ord_id and cluster_name ?

tisnik (approved):
LGTM

juandspy (approved):

PR #1638, opened 2022-06-30
Get rid of `ioutil`

tisnik:
# Description

Get rid of `ioutil`

## Type of change

## Testing steps

Done on CI

## Checklist


consumer/benchmark_test.go
```diff
@@ -16,6 +16,7 @@ package consumer_test
 
 import (
 	"io/ioutil"
```
juandspy, line 18: see line 92, you forgot an ioutil

juandspy (changes requested):
there is a missing ioutil

dpensi (approved):

PR #1643, opened 2022-08-04
setup memory variables

dpensi:
# Description

Set a variable for memory required by pods, the memory will be increased only in production with a PR to app-interface

Fixes hot-fix

## Type of change

## Testing steps

na

## Checklist


tisnik (approved):
LGTM

PR #1653, opened 2022-09-01
[CCXDEV-9196] Attempt to fix Kafka login

matysek:
# Description

Attempt to fix Kafka login - check for cacert 'nil' value

## Type of change


tisnik (changes requested):
strings can't be `nil` in Go

joselsegura (changes requested):
I don't think our current problem in db-writer is related at all with Kafka authentication. According to the logs, we authenticated well, the problem is that there is a missing topic. The library cannot know about missing topic without a proper connection

matysek:
@joselsegura @tisnik thank you for review. Not Go profficient at all just tried if that makes sense.

PR #1657, opened 2022-09-07
attempt to fix stage issue with inconsistent numbers

Bee-lee:
# Description
I'm pushing a part of the fix, I want to see what's so special about stage (*again*)

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/server/rules.go
+++ b/server/rules.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 
+	"github.com/RedHatInsights/insights-operator-utils/generators"
 	"github.com/RedHatInsights/insights-operator-utils/responses"
 	"github.com/RedHatInsights/insights-results-aggregator/storage"
 	"github.com/RedHatInsights/insights-results-aggregator/types"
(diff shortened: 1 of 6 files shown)


dpensi (approved):

juandspy (approved):

server/rules.go
```diff
@@ -210,14 +211,47 @@ func (server HTTPServer) listOfDisabledClusters(writer http.ResponseWriter, requ
 	}
 }
 
-// getFeedbackAndTogglesOnRules
+// getRuleToggleMapForCluster retrieves list of disabled rules and returns a map of rule_ids
```
tisnik, line 214: technically it would be a set if you use `struct{}` instead of `true`

tisnik (approved):
ok

server/rules.go
Bee-lee: @tisnik if it works, I'm planning to modify it to use the empty interface like you showed us some time ago, this was simply faster without modifying more code, is it ok with you?

PR #1749, opened 2023-05-29
[CCXDEV-11124] Add deployment for cache-writer (just Redis)

juandspy:
# Description

Adding a StatefulSet + services for deploying Redis as part of the cache-writer app. It's still missing the deployment of the cache-writer, as it's not developed yet.

It's heavily inspired and copied from https://github.com/openshift-assisted/assisted-events-stream/blob/master/openshift/template.yaml.

Part of #CCXDEV-11124

## Type of change

## Testing steps

Following the docs I added in this same PR, deploying the Redis in an ephemeral namespace and testing manually that it can be queried.

Note that you have to hardcode the password to a specific value because the secret cannot be pulled yet, it's waiting for: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/69387

## Checklist

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -41,6 +41,11 @@ redis-cli -h cache-writer-redis -p 6379 SET mykey "Hello\nWorld";
 redis-cli -h cache-writer-redis -p 6379 GET mykey;
 ```
 
+You can also check that the metrics are exposed:
+```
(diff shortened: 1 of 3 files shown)


Bee-lee (approved):
trust you:)

epapbak (approved):

PR #1773, opened 2023-06-08
[CCXDEV-11216] Add clowdapp for the cache writer

juandspy:
# Description

Add a clowdapp for the cache writer.

## Type of change

## Testing steps

Ephemeral.

## Checklist

### DIFF
--- a/deploy/cache-writer.yaml
+++ b/deploy/cache-writer.yaml
@@ -19,6 +19,154 @@ metadata:
   name: ccx-cache-writer
 objects:
 # TODO: Add the cache-writer deployment + services
+- apiVersion: cloud.redhat.com/v1alpha1
+  kind: ClowdApp
+  metadata:
+    name: ccx-cache-writer
+  spec:
+    envName: ${ENV_NAME}
+    testing:
+      iqePlugin: ccx
+    deployments:
+      - name: db-writer
+        minReplicas: ${{CACHE_WRITER_REPLICAS}}
+        webServices:
+          public:
+            enabled: false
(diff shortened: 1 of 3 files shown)


deploy/cache-writer.yaml
```diff
@@ -19,6 +19,154 @@ metadata:
...
+            - name: INSIGHTS_RESULTS_AGGREGATOR__SERVER__AUTH
+              value: "${AUTH}"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__SERVER__AUTH_TYPE
+              value: "xrh"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__PROCESSING__ORG_WHITELIST_FILE
+              value: "/data/org_whitelist.csv"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__METRICS__ENABLED
+              value: "true"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__STORAGE__TYPE
+              value: redis
```
tisnik, line 79: ok to not use quotes?

tisnik (approved):
LGTM, just one comment

deploy/cache-writer.yaml
juandspy: yep, this kind of variables are automatically converted into strings. You just need to use quotes with numbers or booleans, which would be converted to non-string types. 

If you are not convinced yet, I already tested on ephemeral, so I checked the types are correctly parsed :smile:

epapbak (approved):

PR #1774, opened 2023-06-09
Lower log level to debug

juandspy:
# Description

Lower log level to debug because I cannot understand why we are not receiving messages in stage.

## Type of change

- Configuration update

## Checklist


tisnik (approved):
ok

PR #1782, opened 2023-06-13
[cache-writer][debug] Use same Kafka config as db-writer

juandspy:
# Description

The cache-writer doesn't consume from Kafka. Let's use the same configuration as the db-writer and see what's going on.

## Type of change

- Configuration change

## Checklist


tisnik (approved):
ok

PR #1789, opened 2023-06-14
Add clowder mapping

juandspy:
# Description

Stage is not working because the Kafka is managed and there are some Clowder tricks there. We need to include the `kafkatopics` section in the new app too: https://redhat-internal.slack.com/archives/C022YV4E0NA/p1686731756379229

## Type of change

## Checklist


tisnik (approved):
ok

Bee-lee (approved):

PR #1797, opened 2023-07-20
[CCXDEV-11194] Ignore minimum 3 replicas alert

juandspy:
# Description

AppSRE has an automated alert system that asked us to increase the replicas to up to 3. We don't need so many replicas as our services are really stable and have more than enough resources.

## Type of change

- Configuration update

## Testing steps

None.

## Checklist


tisnik (approved):
ok

PR #1813, opened 2023-08-09
switch aggregator/db-writer in clowder definitions to check for env var ovewriting

Bee-lee:
# Description
I have a suspicion that the env vars are overwriting each other for the two different specs. We probably wouldn't notice because the shared config options are the same (DB etc) for both services.

Although it sounds weird to me, as they should be deployed separately, I can't find any other place where the cloudwatch log stream name is set.

If this confirms my suspicion, we would have to create a separate clowdapp.yaml and new entry in app-interface as we do for cache-writer.

Fixes # (issue)

## Type of change

- Configuration update

## Testing steps
n/a

## Checklist

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -43,116 +43,6 @@ objects:
     testing:
       iqePlugin: ccx
     deployments:
-      - name: aggregator
-        minReplicas: ${{MIN_REPLICAS}}
-        webServices:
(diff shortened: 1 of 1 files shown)


tisnik (approved):
hmm TBH I must trust you

joselsegura (approved):

epapbak (approved):

PR #1814, opened 2023-08-10
[CCXDEV-11471] Upgrade the postgres database that is used in ephemeral

juandspy:
# Description

We are planning on migrating from Postgres v11 to v15. Let's first update the version that clowder uses for ephemeral deployments. This will tell us whether our iqe tests pass for this version.

This won't affect stage nor prod as this `version` is just used for ephemeral.

Part of #CCXDEV-11471

## Type of change
- Configuration update

## Checklist


tisnik (approved):
ok, let' see

Bee-lee (approved):

PR #1833, opened 2023-09-05
Remove kafka_offset mechanism from code

joselsegura:
# Description

The kafka_offset mechanism is not filtering any message, as its only function is to print a warning message. In addition, the mechanism is not implemented correctly as it can only work on single-partition topics and it relies in offset being always incresing, but it can be reset for many reasons.

Fixes #[CCXDEV-11690](https://issues.redhat.com/browse/CCXDEV-11690)

## Type of change

## Testing steps

Regular local tests

## Checklist


tisnik (approved):
LGTM

Bee-lee (approved):

PR #1840, opened 2023-09-12
Fix valid empty reports not being processed

epapbak:
# Description

Fix bug that was introduced while refactoring the code. It is important to process the *valid* empty reports, as they indicate a cluster where detected issues are no longer happening.

## Type of change

## Testing steps

Updated UTs that would otherwise fail as they were using a valid empty report

## Checklist


tisnik (approved):
ok

juandspy (approved):

PR #1845, opened 2023-09-14
CI: don't shallow clone in BDD test stage

epapbak:
# Description

version info is not obtained with `git describe --always --tags --abbrev=0`. So it can either be a tag or a full commit SHA, but not an empty string.

## Type of change

Fix for CI

## Testing steps

CI

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # retrieve the latest tag set in repository
-version=$(git describe --tags --abbrev=0)
+version=$(git describe --always --tags --abbrev=0)
 
 buildtime=$(date)
 branch=$(git rev-parse --abbrev-ref HEAD)


Bee-lee (approved):
maybe passing `--always` arg to `git describe` could also fix it, but this repo is not huge, so the deep clone won't slow things down anyway

epapbak:
> maybe passing `--always` arg to `git describe` could also fix it, but this repo is not huge, so the deep clone won't slow things down anyway

it fixes the `git describe` command, but we still would get an output where the version is not a proper tag. It would be a commit

tisnik (approved):
ok

tisnik (approved):
ok

epapbak:
not needed since https://github.com/RedHatInsights/insights-behavioral-spec/pull/491

PR #1848, opened 2023-09-19
Using a different base image for the final image

joselsegura:
# Description

Using ubi-micro instead ubi-minimal to remove as much dependencies as we can to avoid CVEs.

## Type of change

- Bump-up dependent library (no changes in the code)
- Security fix in dependent library (no changes in the code)
- Security fix in dependent library (changes made in the code)

## Testing steps

Regular CI

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN umask 0022 && \
     make build && \
     chmod a+x insights-results-aggregator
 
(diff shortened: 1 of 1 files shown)


tisnik (approved):
PTAL

epapbak (approved):

Bee-lee (approved):

PR #1852, opened 2023-09-25
Drop support for `jwt` auth type

Bee-lee:
# Description
- `jwt` auth type is not used in any real environment, all environments use `xrh` auth type and using JWT is not supported at all.
- aggregator already uses `xrh` auth type in most UTs

Fixes CCXDEV-9369

## Type of change

## Testing steps
`make before_commit`

## Checklist


tisnik (approved):
ok, but probably sec/ check will say something about key ;)

Bee-lee:
@tisnik same thing is used in smart-proxy and the linters don't seem to care :)

Bee-lee:
@tisnik that doesn't match, the tokens are in https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/server_test.go

epapbak (approved):

PR #1893, opened 2023-11-30
New storage configuration options

tisnik:
# Description

New storage configuration options

## Type of change

- Documentation update

## Testing steps

To be checked on CI.

## Checklist

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -33,6 +33,15 @@ sqlite_datasource = "./aggregator.db"
 log_sql_queries = true
 type = "sql"
 
+[dvo_recommendations_storage]
+db_driver = "sqlite3"
+sqlite_datasource = "./dvo.db"
+log_sql_queries = true
+type = "sql"
+
+[storage_backend]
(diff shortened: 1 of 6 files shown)


matysek (approved):
/lgtm

config.toml
```diff
@@ -33,6 +33,15 @@ sqlite_datasource = "./aggregator.db"
 log_sql_queries = true
 type = "sql"
 
+[dvo_recommendations_storage]
+db_driver = "sqlite3"
+sqlite_datasource = "./dvo.db"
```
Bee-lee, line 38: IMHO I don't think it makes much sense to support sqlite. It isn't supported at all for the "old" storage. In fact you can't use it as migrations will fail

Bee-lee (approved):
LGTM apart of the sqlite support

config.toml
tisnik: removed

juandspy (approved):

PR #1901, opened 2023-12-11
add support for DVO specific migrations

Bee-lee:
# Description
- `migration` module is now more generic and is decoupled from OCP related migrations
- `migration` module doesn't rely on `storage` module
- OCP related migrations were split into an independent module
- DVO related migrations also have their own module (real migrations will come in future PRs)
- depending on new config option `storage_backend.use`, either OCP or DVO migrations will be applied

NOTE: `dvo` databases are not yet created in RDS, we're waiting for AppSRE ([ticket](https://issues.redhat.com/browse/APPSRE-8618))

Fixes # (issue)

## Type of change

## Testing steps
locally with properly created `dvo` and `aggregator` databases. based on `storage_backend.use`, service will properly select the correct migrations

## Checklist


aggregator.go
```diff
@@ -190,7 +189,7 @@ func prepareDBMigrations(dbStorage storage.OCPRecommendationsStorage) int {
 // prepareDB function opens a connection to database and loads all available
 // rule content into it.
 func prepareDB() int {
-	// TODO: when migrations for DVO will be available, update the code below
+	// TODO: when aggregator supports both storages at once, update the code below
```
juandspy, line 192: is there a task to support both storages? I would reference it here just so we know there's a plan

aggregator.go
```diff
@@ -112,29 +112,27 @@ func fillInInfoParams(params map[string]string) {
 // createStorage function initializes connection to preconfigured storage,
 // usually SQLite, PostgreSQL, or AWS RDS.
 func createStorage() (storage.OCPRecommendationsStorage, storage.DVORecommendationsStorage, error) {
-	storageCfg := conf.GetOCPRecommendationsStorageConfiguration()
-	redisCfg := conf.GetRedisConfiguration()
```
tisnik, line 116: is not this one needed for cache-writer

aggregator.go
Bee-lee: yeah, [this one](https://issues.redhat.com/browse/CCXDEV-12316) I created yesterday. good idea

tisnik: yup, look at Dan's tasks

aggregator.go
```diff
@@ -374,14 +374,28 @@ func printEnv() int {
 // getDBForMigrations function opens a DB connection and prepares the DB for
 // migrations. Non-OK exit code is returned as the last return value in case
 // of an error. Otherwise, database and connection pointers are returned.
-func getDBForMigrations() (storage.OCPRecommendationsStorage, *sql.DB, int) {
+func getDBForMigrations() (storage.Storage, *sql.DB, int) {
 	// use OCP recommendations storage only, unless migrations will be available for other storage(s) too
-	db, _, err := createStorage()
+	var db storage.Storage
+	var err error
```
tisnik, line 380: not needed

Bee-lee: fixed, ty

aggregator.go
Bee-lee: fixed, thank you, good catch

storage/ocp_recommendations_storage.go
```diff
@@ -242,8 +246,10 @@ type OCPRecommendationsDBStorage struct {
 func NewOCPRecommendationsStorage(configuration Configuration) (OCPRecommendationsStorage, error) {
 	switch configuration.Type {
 	case types.SQLStorage:
+		log.Info().Str("OCP storage type", configuration.Type).Send()
```
epapbak, line 249: "SQL storage type"

We should stop using "OCP", now that DVO is coming in the picture

Bee-lee: it was actually just SQL before, but now we make a difference in the logs between OCP and DVO storage

migration/ocpmigrations/ocp_migrations_test.go
```diff
@@ -0,0 +1,29 @@
...
+	dbStorage := mockStorage.(*storage.OCPRecommendationsDBStorage)
+
+	return dbStorage.GetConnection(), dbStorage.GetDBDriverType(), closer
+}
+
+func prepareDBAndInfo(t *testing.T) (*sql.DB, types.DBDriver, func()) {
+	db, dbDriver, closer := prepareDB(t)
+
+	if err := migration.InitInfoTable(db); err != nil {
+		closer()
```
epapbak, line 24: I think this should always be called, since the connection is established at some point to create the test DB

Bee-lee: @epapbak we don't want to close it after `prepareDBAndInfo` finishes, it returns the closer function which is then used/deferred in all the functions that use this prepare function

(this code and the `actual_migrations_test.go` is a 1:1 copy from the code that was always here)

tisnik (approved):
LGTM

storage/ocp_recommendations_storage.go
epapbak: ah, my bad. it's in the 'NewOCPRecommendationsStorage' function, so this log does make sense. Scratch this comment too =)

epapbak (approved):
LGTM, nice work!

juandspy (approved):

PR #1914, opened 2023-12-18
[CCXDEV-7292] Stop supporting SQLite

epapbak:
# Description

Big PR, I know, but it's actually a pretty simple change: we stop supporting SQLite, and use PostgreSQL everywhere.
- The changes in config and GitHub workflow YAML files were needed in order for the tests to connect properly to the DB both locally and in the CI environment
- The `unit-tests.sh` and `test.sh` scripts were updated to start the postgreSQL docker image before running the tests, which is now needed since a real postgres is expected (improvement to stop it properly and avoid starting it if a postgres database is already running locally can be done in another PR, if needed)
- All the tests now use `ira_helpers.MustGetPostgresStorage` instead of `ira_helpers.MustGetMockStorage`
- The rest of the changes remove the if-else blocks added to the source code to support SQLite

Fixes CCXDEV-7292

## Type of change

- Configuration update

## Testing steps

CI

## Checklist


juandspy (approved):

tisnik (approved):
LGTM, but not sure about removing `os.Clearenv()
	mustLoadConfiguration("./tests/tests")` from tests

Bee-lee (approved):
good one! 👍

PR #1925, opened 2024-01-08
add functionality for multple DB schemas, add support for both storages at once (aggregator case)

Bee-lee:
# Description
- "empty" `storage_backend.use` is now supported -- uses both storagesa t once (aggregator)
- both aggregator commands now support multiple DB schemas
  - `migrate` command and `start-service` command
- existing UTs cover most of the behavior, adding more to cover new functionality
- the internal logic will need to be improved, because the logic relies on the server running (meaning that db-writer, cache-writer and dvo-writer have a server running for no reason)

Fixes # (issue)

## Type of change

- Configuration update

## Testing steps

## Checklist

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -122,23 +122,15 @@ func createStorage() (storage.OCPRecommendationsStorage, storage.DVORecommendati
 	var dvoStorage storage.DVORecommendationsStorage
 	var err error
 
-	// try to initialize connection to storage
-	backend := conf.GetStorageBackendConfiguration().Use
-	switch backend {
-	case types.OCPRecommendationsStorage:
-		ocpStorage, err = storage.NewOCPRecommendationsStorage(ocpStorageCfg)
-		if err != nil {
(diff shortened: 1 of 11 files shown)


joselsegura (approved):

pr_check.sh
```diff
@@ -20,15 +20,17 @@ set -exv
 # Options that must be configured by app owner
 # --------------------------------------------
 APP_NAME="ccx-data-pipeline"  # name of app-sre "application" folder this component lives in
-REF_ENV="insights-production"
+# TODO set it to "insights-stage" once we have dvo-writer/dvo-extractor stage deploy config
```
juandspy, line 23: this line can be removed

juandspy (approved):

pr_check.sh
Bee-lee: discussed on slack: we'll do it in a separate PR

PR #1955, opened 2024-02-09
[CCXDEV-12475] Insert workload recommendation in DB

juandspy:
# Description

This is the code for inserting the workload recommendations (DVO) into the `dvo.dvo_report` table.

Note that the message we receive is a JSON that relates recommendations with namespaces whereas in the DB we are storing namespaces and the number of recommendations each has. This made me use that for loop and create a map of namespaces so that we don't have to change the DB nor the dvo-extractor.

This is how the DB is populated:

```
org_id          INTEGER NOT NULL, -> same as OCP, we get it from the message
cluster_id      VARCHAR NOT NULL, -> same ^
namespace_id    VARCHAR NOT NULL, -> we get it from each workload under workload_recommendations
namespace_name  VARCHAR,          -> same as above, but can be missing
report          TEXT,             -> we store the complete report, even for the workloads from other namespaces
recommendations INTEGER NOT NULL, -> number of recommendations in the complete report
objects         INTEGER NOT NULL, -> number of workloads for this namespace
reported_at     TIMESTAMP,        -> TODO: I'm not sure what to put here
last_checked_at TIMESTAMP,        -> last time we received an archive
```

We need to revisit the `report` column once the feature is working as it's duplicated in each row for the same cluster.

## Type of change

## Testing steps

Locally + CI.

## Checklist

### DIFF
--- a/consumer/dvo_rules_consumer_test.go
+++ b/consumer/dvo_rules_consumer_test.go
@@ -417,21 +417,7 @@ func TestProcessingEmptyMetricsMissingAttributesWithClosedStorage(t *testing.T) 
 	helpers.FailOnError(t, err, "empty report should not be considered an error at HandleMessage level")
 }
 
-func TestProcessingValidDVOMessageEmptyReportWithRequiredAttributesWithClosedStorage(t *testing.T) {
-	// TODO: Unskip
-	t.Skip("not implemented yet")
-	mockStorage, closer := ira_helpers.MustGetPostgresStorageDVO(t, true)
-
-	mockConsumer := dummyDVOConsumer(mockStorage, true)
-	closer()
-
-	err := consumerProcessMessage(mockConsumer, messageReportNoDVOMetrics)
-	assert.EqualError(t, err, "sql: database is closed")
-}
-
 func TestProcessingCorrectDVOMessageWithClosedStorage(t *testing.T) {
-	// TODO: Unskip
-	t.Skip("not implemented yet")
 	mockStorage, closer := ira_helpers.MustGetPostgresStorageDVO(t, true)
 
 	mockConsumer := dummyDVOConsumer(mockStorage, true)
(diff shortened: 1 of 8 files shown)


consumer/dvo_rules_consumer_test.go
```diff
@@ -421,21 +417,7 @@ func TestProcessingEmptyMetricsMissingAttributesWithClosedStorage(t *testing.T)
 	helpers.FailOnError(t, err, "empty report should not be considered an error at HandleMessage level")
 }
 
-func TestProcessingValidDVOMessageEmptyReportWithRequiredAttributesWithClosedStorage(t *testing.T) {
```
epapbak, line 424: did you remove this test on purpose?

juandspy: Yep because if the report is empty we don't touch the DB. I copied this test from the OCP storage but it makes no sense. I realised later :) The behaviour should be tested in `TestProcessingEmptyMetricsMissingAttributesWithClosedStorage`

storage/dvo_recommendations_storage.go
```diff
@@ -249,16 +287,119 @@ func (storage DVORecommendationsDBStorage) WriteReportForCluster(
...
+	orgID types.OrgID,
+	clusterName types.ClusterName,
+	report types.ClusterReport,
+	recommendations []types.WorkloadRecommendation,
+	lastCheckedTime time.Time,
 ) error {
-	// TODO
-	log.Debug().Msg("updating report in database")
+	// Get reported_at if present before deletion
+	reportedAtMap, err := storage.getReportedAtMap(orgID, clusterName)
```
epapbak, line 298: I would move this after line 304 (why query the storage if we're going to return nil and do nothing anyway)

storage/dvo_recommendations_storage.go
```diff
@@ -105,15 +118,42 @@ func newDVOStorage(configuration Configuration) (DVORecommendationsStorage, erro
...
 // tasks necessary for further service operation.
 func (storage DVORecommendationsDBStorage) Init() error {
-	return nil
+	// Read clusterName:LastChecked dictionary from DB.
+	rows, err := storage.connection.Query("SELECT cluster_id, last_checked_at FROM dvo.dvo_report;")
+	if err != nil {
+		return err
+	}
+
+	log.Debug().Msg("executing last_checked_at query")
```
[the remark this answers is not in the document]
juandspy: That would print it once per report, which could be thousands of times. Not sure if we should be generating that amount of logs

storage/dvo_recommendations_storage.go
juandspy: TY!

epapbak (approved):

Bee-lee (approved):

storage/dvo_recommendations_storage.go
```diff
@@ -105,15 +118,42 @@ func newDVOStorage(configuration Configuration) (DVORecommendationsStorage, erro
...
+			return err
+		}
+
+		storage.clustersLastChecked[clusterName] = lastChecked
+	}
+
+	// Not using defer to close the rows here to:
+	// - make errcheck happy (it doesn't like ignoring returned errors),
+	// - return a possible error returned by the Close method.
+	return rows.Close()
```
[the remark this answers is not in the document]
epapbak: then we can consider it battle tested =)

PR #1961, opened 2024-02-15
Support for multiple broker addresses

epapbak:
# Description

This is the simplest change required to support multiple broker addresses (without refactoring common code out of the service and so on).

Now the configuration of the Kafka Broker expects the field `addresses` instead of `address`, and said field is a comma-separated list of broker addresses. For example: `kafka:9092,kafka:9093,localhost:29092`. The list does not have a minimum or maximum number of addresses, so it can be used as before with only one address: `kafka:29092`.

Unrelated to PR, I enabled an option in golangci-lint so it only fails if new issues are detected (if not, CI wouldn't pass)

Fixes CCXDEV-12462

## Type of change

- Documentation update
- Configuration update

## Testing steps

- UTs
- BDDs

## Checklist

### DIFF
--- a/conf/configuration.go
+++ b/conf/configuration.go
@@ -291,6 +291,42 @@ func loadAllowlistFromCSV(r io.Reader) (mapset.Set, error) {
 	return allowlist, nil
 }
 
+func updateBrokerCfgFromClowder(configuration *ConfigStruct) {
+	// make sure broker(s) are configured in Clowder
+	if len(clowder.LoadedConfig.Kafka.Brokers) > 0 {
+		configuration.Broker.Addresses = ""
+		for _, broker := range clowder.LoadedConfig.Kafka.Brokers {
+			if broker.Port != nil {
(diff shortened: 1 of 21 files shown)


conf/configuration.go
```diff
@@ -355,7 +362,7 @@ func updateStorageConfFromClowder(conf *storage.Configuration) {
 	conf.PGPassword = clowder.LoadedConfig.Database.Password
 }
 
-func updateTopicsMapping(c *ConfigStruct) error {
```
juandspy, line 358: is it safe to remove this error?

juandspy (approved):

conf/configuration.go
epapbak: I think we added it by habit. The function does not seem to do anything that can return an error, so yes, it's safe IMO.

joselsegura (approved):

Bee-lee (approved):

PR #1965, opened 2024-02-20
[CCXDEV-12529][dvo-writer] Use DELETE -> INSERT instead of upsert

juandspy:
# Description

This fix the situation where

1. you have a cluster123, the cluster has 2 namespaces (namespace_A, namespace_B)
2. we process it, store it, when someone requests it, we display that cluster123 has namespace_A and namespace_B
3. in a few hours, we get another archive from cluster123, but this time, it ONLY has namespace_A because the other namespace got finished/deleted/whatever
4. since we were only doing UPDATE .. ON CONFLICT (org_id, cluster_id, namespace_id), this means that namespace_A will be updated, but namespace_B is still going to be in the database
5. when someone requests the data, we're still going to be showing both namespace_A and namespace_B  for 30 days (until next cleanup) or until the cluster disappears from the AMS API

with these changes we make sure outdated reports aren't stored.

## Type of change

## Testing steps

Unit tests.

## Checklist

### DIFF
--- a/consumer/dvo_rules_consumer_test.go
+++ b/consumer/dvo_rules_consumer_test.go
@@ -446,3 +446,24 @@ func TestProcessingDVOMessageWithWrongDateFormatReportNotEmpty(t *testing.T) {
 		))
 	}
 }
+
+func TestDVOKafkaConsumerMockOK(t *testing.T) {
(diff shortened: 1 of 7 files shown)


Bee-lee (approved):
LGTM overall, ty!

epapbak (approved):

PR #1969, opened 2024-02-25
add DVO related endpoints

Bee-lee:
# Description
adds 2 endpoints used for DVO workloads recommendations

- `DVOWorkloadRecommendations = "organization/{organization}/workloads"` is used to get all workloads/namespaces for a given organization

- `DVOWorkloadRecommendationsSingleNamespace = "organization/{organization}/namespace/{namespace}/cluster/{cluster}/workloads"` is used for the single namespace DVO report

TODO:
- more UT coverage of the REST API
- OpenAPI spec
- move types to shared repo

Fixes # (issue)

## Type of change

- Changes in REST API schema
- Behavioral tests (no changes in the code)

## Testing steps
along with [smart-proxy](https://github.com/RedHatInsights/insights-results-smart-proxy/pull/1230), will be done in ephemeral

## Checklist


storage/dvo_recommendations_storage.go
```diff
@@ -411,6 +421,118 @@ func (storage DVORecommendationsDBStorage) getReportedAtMap(orgID types.OrgID, c
...
+		workloads = append(workloads, dvoReport)
+		count++
+	}
+
+	log.Debug().Int(orgIDStr, int(orgID)).Msgf("ReadWorkloadsForOrganization processed %d rows in %v", count, time.Since(tStart))
+
+	return workloads, err
+}
+
+// ReadWorkloadsForClusterAndNamespace returns a single result from the dvo.dvo_report table
```
juandspy, line 479: this function shares a lot of code with `ReadWorkloadsForOrganization`, but probably it's not worth it to refactor them given the benefit we would get

juandspy (approved):

server/dvo_handlers.go
```diff
@@ -0,0 +1,298 @@
...
+		return
+	}
+
+	for _, recommendation := range dvoReport.WorkloadRecommendations {
+		filteredObjects := make([]DVOObject, 0)
+		for i := range recommendation.Workloads {
+			object := &recommendation.Workloads[i]
+
+			// filter out other namespaces
+			if object.NamespaceUID != processedWorkloads.Namespace.UUID {
```
epapbak, line 270: probably not needed, since we're already doing this filtering in the SQL query. Or am I misunderstanding something?

epapbak (approved):

server/dvo_handlers.go
Bee-lee: yeah, this is filtering the objects (workloads) themselves inside each recommendation. Because the whole report is not grouped by namespace.. the whole report always contains data for all namespaces that we have to filter out since we only want 1 namespace. This is something we could refactor to only store the relevant part of the report but for now (until GA) it's not a bad idea to be storing the complete report

PR #1973, opened 2024-02-27
fix wrongly encoded JSON string from the dvo.dvo_report.report column

Bee-lee:
# Description
currently, when you retrieve the `dvo.dvo_report.report` column containing the full JSON report, the string is wrongly encoded and some double-escaping happens _somewhere_ along the way. This needs fixing, but it's not immediately apparent where/when this happens. 

there are 2 issues with the whole `report` string in the DB:
- the escape characters before the quotation marks around the JSON keys/values are double-escaped (`\\\"`)
- the whole JSON string is also wrapped in quotation marks, this time they are only escaped once (`\"{}\"`)

this means that the function `ProcessSingleDVONamespace` works with a "JSON" string that looks like this
`\"{\\\"system\\\":{\\\"metadata\\\":{}}}\"`

This PR fixes `ProcessSingleDVONamespace` to be able to work with the current JSON, as well as ensuring it will be able to work with a correct JSON `{"system":{"metadata":{}}}` -- there are 2 UTs testing both cases

Fixes # (issue)

## Type of change

## Testing steps
locally

## Checklist

### DIFF
--- a/server/dvo_handlers.go
+++ b/server/dvo_handlers.go
@@ -217,7 +217,7 @@ func (server *HTTPServer) getWorkloadsForNamespace(writer http.ResponseWriter, r
 		return
 	}
 
-	processedWorkload := server.processSingleDVONamespace(workload)
+	processedWorkload := server.ProcessSingleDVONamespace(workload)
 
 	log.Info().Uint32(orgIDStr, uint32(orgID)).Msgf(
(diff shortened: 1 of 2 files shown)


juandspy (approved):

epapbak (approved):
ok

PR #1990, opened 2024-03-13
Fix linters Action

juandspy:
# Description

Remove go1.18 errcheck

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -9,7 +9,6 @@ jobs:
     strategy:
       matrix:
         go-version:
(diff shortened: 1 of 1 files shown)


epapbak (approved):

Bee-lee (commented):
we should probably stop building with 1.18 if we're disabling all these checks
https://github.com/RedHatInsights/insights-results-aggregator/blob/master/go.mod#L3
https://github.com/RedHatInsights/insights-results-aggregator/blob/master/Dockerfile#L15

PR #2026, opened 2024-06-24
Fix wrong creation of impactedSince map

epapbak:
# Description

Remove the `LIMIT 1` from the select query. I think what was intended is to select only distinct results, which is not needed neither as the rows have a unique combination `org ID` + `cluster ID` + `rule FQDN` + `error key`.

With this fix, the `impactedSince` map contains an entry for all the rows that exist in the `recommendation `table for a given `org ID + cluster ID`, which ensures the timestamp is conserved for a previously hitting rule.

Fixes CCXDEV-10244

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/storage/ocp_recommendations_storage_test.go
+++ b/storage/ocp_recommendations_storage_test.go
@@ -1186,6 +1186,42 @@ func TestDBStorageInsertRecommendationsNoRuleHit(t *testing.T) {
 	helpers.FailOnError(t, err)
 }
 
+func Test_CCXDEV_10244_DBStorageGetImpactedSinceMap(t *testing.T) {
+	mockStorage, closer := ira_helpers.MustGetPostgresStorage(t, true)
+	defer closer()
+
+	err := mockStorage.WriteRecommendationsForCluster(
+		testdata.OrgID, testdata.ClusterName, testdata.Report3Rules, RecommendationCreatedAtTimestamp,
+	)
+	helpers.FailOnError(t, err)
+
+	// let's retrieve the stored data and ensure it is what we want to work with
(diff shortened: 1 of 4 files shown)


JiriPapousek (approved):

matysek (approved):
lgtm

storage/ocp_recommendations_storage_test.go
```diff
@@ -1186,6 +1186,42 @@ func TestDBStorageInsertRecommendationsNoRuleHit(t *testing.T) {
...
+	//We have 3 recommendations with different rule FQDN for this Org ID + Clustername
+
+	//query used before fix.
+	query := "SELECT rule_fqdn, error_key, created_at FROM recommendation WHERE org_id = $1 AND cluster_id = $2 LIMIT 1;"
+	rkcMap, err := storage.GetRuleKeyCreatedAtMap(mockStorage.(*storage.OCPRecommendationsDBStorage), query, testdata.OrgID, testdata.ClusterName)
+	helpers.FailOnError(t, err)
+	assert.Equal(t, len(rkcMap), 1)
+
+	//query used after fix, without limiting number of rows retrieved from the DB
+	query = "SELECT rule_fqdn, error_key, created_at FROM recommendation WHERE org_id = $1 AND cluster_id = $2;"
```
juandspy, line 1219: I'd use a `const` for this query so that if we change the original one in the future, we realise it could break this feature

Bee-lee (approved):
good catch

storage/ocp_recommendations_storage_test.go
epapbak: You're right. I was refactoring a little bit outside of the PR, but I added it here instead. See https://github.com/RedHatInsights/insights-results-aggregator/pull/2026/commits/685049b7648caa251faa379aaf6836d285b697f3

PR #2031, opened 2024-07-10
active cluster list filtering in `DVOWorkloadRecommendations` endpoint

Bee-lee:
# Description
This PR implements a cluster filtering mechanism in the DVO `DVOWorkloadRecommendations` endpoint. Based on [benchmarks](https://gist.github.com/Bee-lee/1ef7ea86db29b41ef81443964015679e), the absolute fastest way is to pass the cluster list as a JSON-encoded array, even though we need to have them as a map in the end, it's much faster than marshaling a map of equal size. Iterating over maps takes longer than iterating over slices, but more importantly `json` takes much longer to marshal/unmarshal a map than a slice (5-10x longer at worst).

- `organization/{org_id}/workloads` now accepts a `POST` method. If a `POST` request is made, the handler expects a JSON-encoded cluster UUID list in the request body. Failure to provide the body results in a `400 Bad Request`. 
- by making a `POST` request, the cluster filtering is enabled, meaning that if a valid empty list is provided `{[]}`, the handler will filter every result out, returning an empty results set. 
- support for existing `GET` method is kept, cluster filtering is disabled unless a `POST` request is made
 
Fixes https://issues.redhat.com/browse/CCXDEV-12957

## Type of change

- Changes in REST API schema
- REST API tests

## Testing steps

## Checklist

### DIFF
--- a/server/router_utils.go
+++ b/server/router_utils.go
@@ -221,7 +221,7 @@ func ReadClusterMapFromBody(writer http.ResponseWriter, request *http.Request) (
 		return
 	}
 
-	// decode from body
+	// decode cluster list from body
 	var clustersInRequest []types.ClusterName
 	err := json.NewDecoder(request.Body).Decode(&clustersInRequest)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


ikerreyes (approved):

juandspy (approved):

server/router_utils.go
```diff
@@ -206,3 +206,34 @@ func getRuleAndErrorKeyFromRuleID(ruleIDWithErrorKey string) (
...
+	}
+
+	// decode cluster list from body
+	var clustersInRequest []types.ClusterName
+	err := json.NewDecoder(request.Body).Decode(&clustersInRequest)
+	if err != nil {
+		handleServerError(writer, err)
+		return
+	}
+
```
epapbak, line 231: "by making a POST request, the cluster filtering is enabled, meaning that if a valid empty list is provided {[]}, the handler will filter every result out, returning an empty results set."

What about, if the length of the received list is 0, returning the empty results set directly? We'd save some processing that way.

juandspy (approved):

PR #2059, opened 2024-09-30
Use Redis sets instead of strings

epapbak:
# Description

Using Redis strings gets real slow when the number of entries starts growing. We have around 750000 values in the cache on any given day, which is not a lot, but retrieving values from there takes up to 30 seconds sometimes, depending on what's happening in the Redis server. Not viable. This PR replaces the Reds strings with Sets, which is way more efficient in this case.

Fixes CCXDEV-13117

## Type of change

## Testing steps

CI

## Checklist


epapbak:
Not valid solution, as we lose the TTL per entry

PR #2065, opened 2024-10-01
Fix linters and detected issues

epapbak:
# Description

Fixed gosec and ineffassign as well as the issues detected once they were working properly. Most of the UTs changes are due to handling orgID as a uint32 where it was defined/converted as a uint64.

Fixes broken CI

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -224,8 +224,8 @@ type Namespace struct {
 
 // DVOMetadata structure contains basic information about workload metadata
 type DVOMetadata struct {
-	Recommendations int         `json:"recommendations"`
-	Objects         int         `json:"objects"`
+	Recommendations uint        `json:"recommendations"`
+	Objects         uint        `json:"objects"`
 	ReportedAt      string      `json:"reported_at"`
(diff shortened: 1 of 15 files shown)


types/types.go
```diff
@@ -224,8 +224,8 @@ type Namespace struct {
 
 // DVOMetadata structure contains basic information about workload metadata
 type DVOMetadata struct {
-	Recommendations int         `json:"recommendations"`
-	Objects         int         `json:"objects"`
+	Recommendations uint        `json:"recommendations"`
+	Objects         uint        `json:"objects"`
```
matysek, line 228: Just curious - might this change affect the data and they format received from kafka or the data stored/read from DB?

epapbak: Data moved through Kafka is not affected, as the JSON remains the same. This indicates at Go level that the integer is unsigned (so always positive). Same thing with the data stored in the DB, it is stored in the format specified in the database (INTEGER NOT NULL) as long as the data is a valid integer, which should not change.

juandspy (approved):

PR #2080, opened 2024-11-04
[CCXDEV-14562] New table for runtimes heartbeats

ikerreyes:
A migration included in the DVO ones (for simplicity) creates a new table to store runtimes heartbeats data (basically ID and timestamp).

Included in the DVO ones to avoid creating new migration schema and another entry point. In addition, this table will be populated by DVO-writer (a modified version of the service) and updated by a new service.

### DIFF
--- a/migration/dvomigrations/mig_0005_runtimes_heartbeats_last_check.go
+++ b/migration/dvomigrations/mig_0005_runtimes_heartbeats_last_check.go
@@ -10,7 +10,7 @@ import (
 var mig0005CreateRuntimesHeartbeats = migration.Migration{
 	StepUp: func(tx *sql.Tx, _ types.DBDriver) error {
 		_, err := tx.Exec(`
-			CREATE TABLE runtimes.heartbeats (
+			CREATE TABLE dvo.runtimes_heartbeats (
 				instance_id     VARCHAR NOT NULL,
 				last_checked_at TIMESTAMP,
 				PRIMARY KEY(instance_id)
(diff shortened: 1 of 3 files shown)


migration/dvomigrations/mig_0005_runtimes_heartbeats_last_check.go
```diff
@@ -0,0 +1,36 @@
...
+	"database/sql"
+
+	"github.com/RedHatInsights/insights-results-aggregator/migration"
+	"github.com/RedHatInsights/insights-results-aggregator/types"
+)
+
+var mig0005CreateRuntimesHeartbeats = migration.Migration{
+	StepUp: func(tx *sql.Tx, _ types.DBDriver) error {
+		_, err := tx.Exec(`
+			CREATE TABLE runtimes.heartbeats (
```
ikerreyes, line 14: Not sure if we are allowed to create tables with this naming, or we need to stick to dvo ones.

juandspy (changes requested):
please add some unit tests: https://github.com/RedHatInsights/insights-results-aggregator/blob/master/migration/dvomigrations/actual_migrations_test.go

ikerreyes:
> please add some unit tests: https://github.com/RedHatInsights/insights-results-aggregator/blob/master/migration/dvomigrations/actual_migrations_test.go

Is it worth for a PoC?

migration/dvomigrations/mig_0005_runtimes_heartbeats_last_check.go
ikerreyes: Seems that is not allowed: `schema \"runtimes\" does not exist"`

juandspy:
> > please add some unit tests: https://github.com/RedHatInsights/insights-results-aggregator/blob/master/migration/dvomigrations/actual_migrations_test.go
> 
> Is it worth for a PoC?

I think so. I'd leave BDD and IQE tests apart, but for Go repos UTs are a must IMO. It's a really simple way of making it easier for reviewers

ikerreyes:
> > > please add some unit tests: https://github.com/RedHatInsights/insights-results-aggregator/blob/master/migration/dvomigrations/actual_migrations_test.go
> > 
> > 
> > Is it worth for a PoC?
> 
> I think so. I'd leave BDD and IQE tests apart, but for Go repos UTs are a must IMO. It's a really simple way of making it easier for reviewers

IMO it is wasted effort as we know this code is never going to be reused and will be dropped. But I will stick to your policies.

migration/dvomigrations/mig_0005_runtimes_heartbeats_last_check.go
ikerreyes: Sticked to dvo ones for simplicity

juandspy (approved):

PR #2088, opened 2024-11-07
[CCXDEV-14563] Runtimes results filtering

ikerreyes:
Filter runtimes (DVO as we are reusing those) reports based on heartbeats information last checked
information.


storage/dvo_recommendations_storage.go
```diff
@@ -597,11 +600,113 @@ func (storage DVORecommendationsDBStorage) ReadWorkloadsForClusterAndNamespace(
...
+		return dvoReport, nil
+	}
+
+	processedReport := dvoReport.Report
+	processedReport = strings.ReplaceAll(processedReport, "\\n", "")
+	processedReport = strings.ReplaceAll(processedReport, "\\", "")
+	processedReport = processedReport[1 : len(processedReport)-1]
+
+	// filter report
+	var reportData types.DVOMetrics // here we will miss part of the original report, would it matter?
```
juandspy, line 648: > // here we will miss part of the original report, would it matter?

I don't think so: the fields the UI cares about are the ones defined in the struct, so we should be fine.

storage/dvo_recommendations_storage.go
```diff
@@ -597,11 +600,113 @@ func (storage DVORecommendationsDBStorage) ReadWorkloadsForClusterAndNamespace(
...
+	// Begin a new transaction.
+	tx, err := storage.connection.Begin()
+	if err != nil {
+		return err
+	}
+
+	err = func(tx *sql.Tx) error {
+		// Check if there is a more recent report for the cluster already in the database.
+		_, err := tx.Exec(
+			"INSERT INTO dvo.runtimes_heartbeats VALUES ($1, $2);", // ON DUPLICATE KEY UPDATE last_checked_at = VALUES(last_checked_at)
```
juandspy, line 700: > `ON DUPLICATE KEY UPDATE last_checked_at = VALUES(last_checked_at)`

is this a left over?

Bee-lee (commented):
how much data are we expecting to be processing? could it affect the existing DB/traffic?

Bee-lee:
I don't see a DB migration creating the `runtimes_hearbeats` table

storage/dvo_recommendations_storage.go
ikerreyes: Just for myself, it is a valid postgresql syntax, but I am getting a DB error when using it.

storage/dvo_recommendations_storage_test.go
```diff
@@ -807,3 +808,34 @@ func TestDVOStorageReadWorkloadsForNamespace_MissingData(t *testing.T) {
...
+	// write heartbeats
+	err = mockStorage.WriteHeartbeat("UID-0099", time.Now().UTC())
+	helpers.FailOnError(t, err)
+	err = mockStorage.WriteHeartbeat("UID-0100", time.Now().UTC().Add(-1*time.Hour).UTC())
+	helpers.FailOnError(t, err)
+
+	report, err := mockStorage.ReadWorkloadsForClusterAndNamespace(testdata.OrgID, testdata.ClusterName, "NAMESPACE-UID-A")
+	helpers.FailOnError(t, err)
+
+	assert.Equal(t, uint(1), report.Objects)
```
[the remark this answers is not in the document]
ikerreyes: do we have that check as a function I can reuse? Otherwise, given that that value is set from the code (it won't be modified unless we run the code we want), and given this is a PoC, I think it is enough. But again, personal opinion, if you want more checks, I will add those.

ikerreyes:
> how much data are we expecting to be processing? could it affect the existing DB/traffic?

3 to 4 archives, and then repeated heartbeats for them every few seconds. But heartbeats will go through a new service.

ikerreyes:
> I don't see a DB migration creating the `runtimes_hearbeats` table

It was here https://github.com/RedHatInsights/insights-results-aggregator/pull/2080 To help I have created a dummy MR with the branch we are using for the PoC so we can see the "merged" code at once: https://github.com/RedHatInsights/insights-results-aggregator/pull/2090

storage/dvo_recommendations_storage.go
ikerreyes: Great, thanks. I will update the comment

storage/dvo_recommendations_storage_test.go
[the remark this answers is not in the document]
juandspy: yes, that was exactly what I wanted to know: if we are properly filtering the inactive instances

storage/dvo_recommendations_storage.go
```diff
@@ -56,6 +56,7 @@ type DVORecommendationsStorage interface {
 		string,
 	) (types.DVOReport, error)
 	DeleteReportsForOrg(orgID types.OrgID) error
+	WriteHeartbeat(string, time.Time) error
```
juandspy, line 59: this is not called outside of the tests. Is it intended? I guess for now it is

ikerreyes: Yes for now we do not have modified the writers to start writing heartbeats data.

juandspy (approved):

PR #2093, opened 2024-11-14
[CCXDEV-14564] DVO writer write heartbeats

ikerreyes:
Once a runtimes report is received, the heartbeats table should be populated with instances seen there.

### DIFF
--- a/storage/dvo_recommendations_storage.go
+++ b/storage/dvo_recommendations_storage.go
@@ -57,6 +57,7 @@ type DVORecommendationsStorage interface {
 	) (types.DVOReport, error)
 	DeleteReportsForOrg(orgID types.OrgID) error
 	WriteHeartbeat(string, time.Time) error
+	WriteHeartbeats([]string, time.Time) error
 }
 
 const (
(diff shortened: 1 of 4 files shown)


storage/dvo_recommendations_storage.go
```diff
@@ -723,3 +724,51 @@ func (storage DVORecommendationsDBStorage) WriteHeartbeat(
...
+
+	sqlStr += ";"
+
+	// Begin a new transaction.
+	tx, err := storage.connection.Begin()
+	if err != nil {
+		return err
+	}
+
+	err = func(tx *sql.Tx) error {
```
juandspy, line 756: is this needed? can't you just replace all this block with `_, err := tx.Exec(sqlStr, vals...)`?

storage/dvo_recommendations_storage.go
```diff
@@ -723,3 +724,51 @@ func (storage DVORecommendationsDBStorage) WriteHeartbeat(
 
 	return err
 }
+
+// WriteHeartbeats insert multiple heartbeats
+func (storage DVORecommendationsDBStorage) WriteHeartbeats(
```
juandspy, line 729: wouldn't it make sense to remove `WriteHeartbeat` and replace with `WriteHeartbeats`? in the end they do the same

ikerreyes: Well, my plan was to convert it later to UpdateHeartbeat, but OK. I will remove it.

ikerreyes: Done

Bee-lee (approved):

PR #2098, opened 2024-11-19
[CCXDEV-14564] Heartbeats consumer

ikerreyes:
# Description

Code for a heartbeats consumer that updates the runtimes table entry with new timestamps for the heartbeats. It is a deployment within an existing service to make deployment easier.


consumer/heartbeats_consumer.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/rs/zerolog/log"
+
+	// "github.com/RedHatInsights/insights-results-aggregator/types"
+	"github.com/Shopify/sarama"
+
+	"github.com/RedHatInsights/insights-results-aggregator/broker"
+)
+
+// Metadata taken from https://github.com/redhatinsights/insights-ingress-go/internal/validators/types.go
+type Metadata struct {
```
juandspy, line 23: FYI: you don't need to define all the fields of the JSON, just the ones you are interested in. This way if it ever removes a field you don't care about, your implementation won't be broken :)

consumer/heartbeats_consumer.go
```diff
@@ -0,0 +1,315 @@
...
+	Timestamp   time.Time `json:"timestamp"`
+}
+
+// KkfMessageProcessor Processor for kafka messages used by NewKafkaConsumer
+type KkfMessageProcessor interface {
+	HandleMessage(msg *sarama.ConsumerMessage) error
+}
+
+// KfkConsumer ...
+type KfkConsumer struct {
```
juandspy, line 64: why do you need to write the consumer from scratch? for other writers we are reusing [this one](https://github.com/RedHatInsights/insights-results-aggregator/blob/8ea82653218e805ceb9398140aad17c854c976d4/consumer/kafka_consumer.go#L61)

consumer/heartbeats_consumer.go
```diff
@@ -0,0 +1,315 @@
...
+		MessageProcessor: processor,
+		ready:            make(chan bool),
+	}
+
+	return consumer, nil
+}
+
+// Serve starts listening for messages and processing them. It blocks current thread.
+// TODO pass ctx, cancel := context.WithCancel(context.Background())
+func (consumer *KfkConsumer) Serve(ctx context.Context) {
```
juandspy, line 107: same question as above: why not reusing the [already existing implementation of KafkaConsumer](https://github.com/RedHatInsights/insights-results-aggregator/blob/8ea82653218e805ceb9398140aad17c854c976d4/consumer/kafka_consumer.go#L153)?

consumer/heartbeats_consumer.go
juandspy: f.e.: https://github.com/RedHatInsights/insights-results-aggregator/blob/8ea82653218e805ceb9398140aad17c854c976d4/consumer/dvo_rules_consumer.go#L25

storage/dvo_recommendations_storage.go
```diff
@@ -730,3 +731,25 @@ func (storage DVORecommendationsDBStorage) WriteHeartbeats(
 
 	return err
 }
+
+// UpdateHeartbeat update timestamp of a heartbeat
+func (storage DVORecommendationsDBStorage) UpdateHeartbeat(
```
juandspy, line 736: why do you use `DVORecommendationsDBStorage` instead of the Heartbeats storage?

consumer/heartbeats_consumer.go
ikerreyes: To be more explicit, as I might be wrong, [KafkaConsumer](https://github.com/RedHatInsights/insights-results-aggregator/blob/master/consumer/kafka_consumer.go#L61) requires a [MessageProcessor](https://github.com/RedHatInsights/insights-results-aggregator/blob/master/consumer/kafka_consumer.go#L65C2-L65C18). Its definition of methods require an [incomingMessage](https://github.com/RedHatInsights/insights-results-aggregator/blob/master/consumer/processing.go#L52) which is different than the ingress incoming message format. I did not know how to reconcile those two.

I feel that extending the incoming struct to support both is not the way, and I couldn't workaround it creating a new type to be one or the other. Plus that will require further changes I believe.

consumer/heartbeats_consumer.go
juandspy: no worries, it was just a note for your Go knowledge :D just saying you don't need to define all the fields

consumer/heartbeats_consumer.go
juandspy: ok, ty!

juandspy (approved):

PR #2101, opened 2024-11-22
[CCXDEV-14564] Heartbeats consumer integration

ikerreyes:
Adding all the code to start the heartbeat serice
(making sure DVO DB is ready, setting the consumer and starting it)

Aggregator main code has been modified to accommodate the new service.

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -381,7 +381,7 @@ The commands are:
     print-version-info  prints version info
     migration           prints information about migrations (current, latest)
     migration <version> migrates database to the specified version
-	start-heartbeater   start the heartbeat processor service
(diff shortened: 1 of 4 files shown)


aggregator.go
```diff
@@ -373,6 +381,7 @@ The commands are:
     print-version-info  prints version info
     migration           prints information about migrations (current, latest)
     migration <version> migrates database to the specified version
+	start-heartbeater   start the heartbeat processor service
```
juandspy, line 384: wrong indent?

juandspy (approved):

aggregator.go
ikerreyes: tab instead of spaces

PR #2182, opened 2025-05-05
[CCXDEV-14922] Create Clowder integration for Redis config

JiriPapousek:
# Description

We need to have Clowder integration in place for Redis in order to start using Valkey. That includes modifying the `CreateRedisClient` function in insights-aggregator-utils.

## Type of change

- Configuration update

## Testing steps

Tested locally (and in ephemeral).

## Checklist


tests/config1.toml
```diff
@@ -46,8 +46,10 @@ use = ""
 [redis]
 database = 0
 endpoint = "localhost:6379"
+username = ""
 password = ""
 timeout_seconds = 30
+use_clowder = true
```
ikerreyes, line 52: Why we need this? Whether clowder is enabled or not should be done at runtime, not from the config

deploy/redis-writer.yaml
```diff
@@ -0,0 +1,253 @@
...
+        minReplicas: ${{CACHE_WRITER_REPLICAS}}
+        webServices:
+          public:
+            enabled: false
+          private:
+            enabled: false
+          metrics:
+            enabled: true
+        podSpec:
+          env:
```
ikerreyes, line 44: I would prefer to create a ConfigMap with the configuration from the parameter rather than setting all these many env vars.

conf/configuration.go
```diff
@@ -356,9 +357,30 @@ func updateConfigFromClowder(c *ConfigStruct) error {
 		fmt.Println(noStorage)
 	}
 
+	// get in-memory DB configuration from clowder
+	if !c.Redis.UseClowder {
```
juandspy, line 361: I guess this is needed in order to support both the cache-writer and the velkey-writer being deployed on clowder at the same time, right? I'm asking because we are not using a custom `use_clowder` config for postgres for example

conf/configuration.go
```diff
@@ -356,9 +357,30 @@ func updateConfigFromClowder(c *ConfigStruct) error {
...
+	} else {
+		fmt.Println(noInMemoryDB)
+	}
+
 	return nil
 }
 
+func updateRedisConfig(conf *storage.RedisConfiguration) {
+	conf.RedisEndpoint = fmt.Sprintf("%s:%d", clowder.LoadedConfig.InMemoryDb.Hostname, clowder.LoadedConfig.InMemoryDb.Port)
+	if clowder.LoadedConfig.InMemoryDb.Username != nil {
```
juandspy, line 376: shall we return an error? if the clowder config has no username/password there's definitely something wrong

conf/configuration.go
```diff
@@ -356,9 +357,30 @@ func updateConfigFromClowder(c *ConfigStruct) error {
...
+		conf.RedisUsername = *clowder.LoadedConfig.InMemoryDb.Username
+	}
+	if clowder.LoadedConfig.InMemoryDb.Password != nil {
```
juandspy, line 379: same here

deploy/redis-writer.yaml
```diff
@@ -0,0 +1,253 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: redis-writer
+objects:
+# TODO: Add the cache-writer deployment + services
```
juandspy, line 21: this is probably a leftover

deploy/redis-writer.yaml
```diff
@@ -0,0 +1,253 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: redis-writer
+objects:
```
juandspy, line 20: why don't you add the valkey-writer as a new object inside the cache-writer.yaml? that way you will be able to reuse all the parameters + no need to change app-interface

deploy/redis-writer.yaml
juandspy: @ikerreyes we can also add this to the cache-writer.yaml (https://github.com/RedHatInsights/insights-results-aggregator/pull/2182/files#r2075573069)

tests/config1.toml
juandspy: I think this is because the current cache-writer uses clowder but not the inmemorydb from clowder. In any case, as the cache-writer defines no inmemorydb, we don't need the `use_clowder` but rather a "is inmemorydb in clowder config?"

JiriPapousek: True, I'll remove that flag :) The intention was as Juan described.

conf/configuration.go
JiriPapousek: Actually, ephemeral instance does not set username or password. There's no auth.

deploy/redis-writer.yaml
JiriPapousek: That will be much better, will do :)

deploy/cache-writer.yaml
```diff
@@ -18,7 +18,166 @@ kind: Template
...
+    name: valkey-writer
+    annotations:
+      ignore-check.kube-linter.io/minimum-three-replicas: "This app doesn't have that much traffic"
+  spec:
+    inMemoryDb: true
+    envName: ${ENV_NAME}
+    testing:
+      iqePlugin: ccx
+    deployments:
+      - name: db-writer
```
ikerreyes, line 33: Do we want to reuse this name?

deploy/cache-writer.yaml
```diff
@@ -18,7 +18,166 @@ kind: Template
...
+        minReplicas: ${{CACHE_WRITER_REPLICAS}}
```
ikerreyes, line 34: I think minReplicas is deprecated in clowder spec

ikerreyes (approved):

deploy/cache-writer.yaml
```diff
@@ -18,7 +18,166 @@ kind: Template
...
+              value: "${KAFKA_BOOTSTRAP_HOST}:${KAFKA_BOOTSTRAP_PORT}"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__BROKER__TIMEOUT
+              value: "${KAFKA_TIMEOUT}"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__BROKER__TOPIC
+              value: "${INCOMING_TOPIC}"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__BROKER__DEAD_LETTER_QUEUE_TOPIC
+              value: ""
+            - name: INSIGHTS_RESULTS_AGGREGATOR__BROKER__PAYLOAD_TRACKER_TOPIC
+              value: "${PAYLOAD_TRACKER_TOPIC}"
+            - name: INSIGHTS_RESULTS_AGGREGATOR__BROKER__SERVICE_NAME
```
juandspy, line 54: is it safe to reuse the same service name as cache-writer? TBH I don't know where it is used. I thought we only use the group ID

deploy/cache-writer.yaml
JiriPapousek: Tbh, I don't know where is this name actually used. Pod name is different.

juandspy: it's the deployment name. If you make a `oc get deployments` you will see it's called `ccx-cache-writer-db-writer`, which comes from `{CLOWDAPP_NAME}-{DEPLOYMENT_NAME}`. In this case it will be `valkey-writer-db-writer`.

juandspy: You can even reuse the clowdapp for cache-writer having a `redis-writer` and another `valkey-writer` deployments

juandspy: I mean, instead of having 2 clowdapps in the Template, having 2 deployments in the clowdapp

deploy/cache-writer.yaml
JiriPapousek: I changed it, just to be safe.

deploy/cache-writer.yaml
JiriPapousek: Ah, I see. But I would probably leave it as it is.

deploy/cache-writer.yaml
JiriPapousek: Replaced with `replicas`

juandspy (approved):

PR #2217, opened 2025-06-19
[CCXDEV-15205] Change provider of sarama package

JiriPapousek:
# Description

Change provider of `sarama` package, since it was changed from `Shopify` to `IBM` some time ago (including `insights-operator-utils` library, that uses new version of `sarama`).

## Type of change

- Bump-up dependent library (no changes in the code)
- Configuration update

## Testing steps

NA

## Checklist


aggregator_test.go
```diff
@@ -211,9 +213,17 @@ func TestStartConsumer_BadBrokerAddress(t *testing.T) {
 		"INSIGHTS_RESULTS_AGGREGATOR__BROKER__ENABLED":   "true",
 	})
 
+	// MockBroker does not currently work with newer Kafka versions,
```
ikerreyes, line 216: Do we have/need a Jira to track this?

PR #2524, opened 2026-06-10
[CCXDEV-15434] Old tutorial entries causing a log error report

joselsegura:
Set lower message level for non-error when a disabled rule is not well formed
Cleanup of cluster_rule_togger database, removing entries without error_key.

# Description

Some old rules like the tutorial one, when disabled by an user, they didn't contain an error key. Now, this is causing that every time a report is generated, if a rule without error key is disabled in the database, a log error message is printed and captured by Glitchtip.

This fix attack the problem from 3 fronts:

1. Make the log level warning instead of error, as it is not causing a malfunctioning of the endpoint.
2. Create a new migration in order to perform a clean up of the table, removing all the entries without error key from the database.

Fixes #[CCXDEV-15434](https://redhat.atlassian.net/browse/CCXDEV-15434)

## Type of change

## Testing steps

Created unit tests

## Checklist

[CCXDEV-15434]: https://redhat.atlassian.net/browse/CCXDEV-15434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/migration/ocpmigrations/actual_migrations_test.go
+++ b/migration/ocpmigrations/actual_migrations_test.go
@@ -17,6 +17,7 @@ package ocpmigrations_test
 import (
 	"database/sql"
 	"database/sql/driver"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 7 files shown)


Bee-lee (commented):
did you forget to include the migration file?

ikerreyes (commented):
What will happen if in the future we introduce by some error again empty error keys? Will another migration be needed?

joselsegura:
> What will happen if in the future we introduce by some error again empty error keys? Will another migration be needed?

First, probably we won't notice as the log level was reduced.

What do you suggest? Include some code to avoid inserting that type of error data into the DB?

ikerreyes:
> What do you suggest? Include some code to avoid inserting that type of error data into the DB?

No suggestion. I was just thinking. If we need to do another migration which will be the same as this one sound a bit weird. But if we won't notice, then I think it is fine.

ikerreyes (approved):

Bee-lee (approved):

